PR #8, opened 2022-03-01
Update backup.sh to delete influxdb snapshots after backup is complete

cwagz:
Here is the change to backup.sh to delete the snapshots after the backup is complete.


jasonacox:
Thank you @cwagz !

PR #9, opened 2022-03-01
Update backup.sh

cwagz:
Removed -it from docker exec influxdb snapshot backup line 26. This causes a "The input device is not a TTY" error when running the script via cron or other methods where tty is not available. When this happens, the script will run to completion, but the snapshots are not created.

The snapshot backup appears to work with output to the screen when run via shell without the -it flag so I believe this is not necessary in the instance. Removing the -it will allow the command to complete if ran manually from shell or via cron.


jasonacox:
Awesome catch, @cwagz !  Thank you.

PR #19, opened 2022-04-15
Changes to allow a smooth upgrade experience.

nhasan:


jasonacox:
Thanks @nhasan !  This is an elegant solution. I'll update the README.

PR #20, opened 2022-04-16
Remove code duplication. Fix logic inversion error.

nhasan:


jasonacox:
Thank you!  Nice catch and elegant fix.

PR #50, opened 2022-07-10
Fix Current State panel so it is not affected by different time range selections

mcbirse:
Current State data will always be retrieved from within the past minute only (instead of based on $timeFilter).


jasonacox:
Thanks for this @mcbirse !  Great find and great fix. Also, congratulations on your first pull request!

PR #66, opened 2022-08-17
Remove dew_point from API service of Weather411

mcbirse:
Remove `dew_point` from API service of Weather411 as the data is not available with "weather" API call to OpenWeatherMap.


jasonacox:
Great catch!  Thanks @mcbirse !

Issue #42, opened 2022-06-21
Enhancement: Temperature line on Live Monitoring graph


mp09:
Enhancement: Temperature line on Live Monitoring graph

Temperature is extremely relevant to power usage if you have/use Air Conditioning. It would be great to see a line graph of that on the live monitoring graph as another data input.

mp09:
Seems like the lat/long info could be used to determine this as well (with a weather data source.

jasonacox:
Hi @mp09 - That's definitely possible.  The challenge we run into is that the API endpoints we scrape to get the local temperature based on lat/long tend to require that the user set up a API token and observe rate limiting on the amount of calls allowed.  Grafana may have a plugin that helps.  I'll see if I can figure out a good way to do this more generically, but would love some help.

jasonacox:
I have an idea.  There is a plugin for telegraf to pull OpenWeatherMap current location weather data (including temperature) - see https://github.com/influxdata/telegraf/blob/master/plugins/inputs/openweathermap/README.md. 

* We could make this an option during setup (e.g. - `Do you want to pull current weather data from OpenWeatherMap?`)
* It would require the user to go to https://openweathermap.org/ to set up an account and copy the API key which they would paste into the setup process.
* If weather data is available, the dashboard.json can render it.  Would the preference be to have it show up on the main Live Monitoring -> "Energy Usage" graph or a separate panel?

jasonacox:
Ok, I used the OpenWeatherMap plugin to get the data into InfluxDB and have added it to the "Energy Usage" graph.  Here is an example of adding Temperature (F) and Cloud Coverage (%). No prefect but I plan to experiment with better visuals but looking to see if this is the general direction that would be helpful.

<img width="861" alt="image" src="https://user-images.githubusercontent.com/836718/184048653-09e9f65d-c017-4cae-8a60-5b81bc35966a.png">

jasonacox:
Update: I have created a branch called "weather" that is my running test configuration.  Go [here](https://github.com/jasonacox/Powerwall-Dashboard/tree/weather) if you want to check it out. Keep in mind that it isn't integrated into setup.sh or upgrade.sh yet.  I'm still testing the new container.  

Here is my approach:

* I created a new service called `weather411` that runs in a container and periodically (every 10m) reaches out to [OpenWeatherMap.org](https://openweathermap.org/) to get the latest weather data for your location.
[...]

PR #69, opened 2022-08-23
Fix rain and snow values not being retrieved

mcbirse:
Fix rain and snow values not being retrieved (refer issue #42)


jasonacox:
Good find @mcbirse !  Thank you!

PR #72, opened 2022-08-28
Fix missing space in if statement

mcbirse:
Fix missing space in if statement which causes an error during script execution: `./weather.sh: line 110: [-z: command not found`


jasonacox:
Thanks @mcbirse !

PR #73, opened 2022-09-04
Update dashboards and grafana

techlover1:
Updated dashboards to be compatible with newer versions of grafana, as well as updated docker-compose with the current latest version.

I am unable to test the metrics that depend on tesla inverters/powerwall plus, but everything else seems to be working as expected


jasonacox:
First observations - 

1) The 9.x version menus are different so it will require an update to the README, setup.sh and upgrade.sh instructions for users.

2) It seems the Power Meter panels have lost their color and font size - I'll need to figure out how to bring this back:

Old:
image

New:
image

jasonacox:
This is great work @techlover1 !  I'm testing it now and want to run it for a while before we merge this.  We also still need to fix the "Current State" panel. I see you added updates, but it doesn't seem to fix the clipping.  I'm see what I can figure out as well.

image

Some screenshots...

image

Powerwall+
image
image

techlover1:
I had not yet fixed the current state panel. Turns out it just needed to be a little taller

jasonacox:
Nice!  That did fix it!  @mcbirse how does this look for your setup?

image

jasonacox:
Under the Powerwall+ section, there are a few errors:

image

* _String Voltage_: The "A string" is missing due to a typo on Voltage in the query.
* _String Current_:  Has the wrong unit, it should be ampere (A).
* _String Power_: Has the wrong unit, it should be wattage (W).

jasonacox:
Merge complete.  Thank you @techlover1 !

Issue #80, opened 2022-09-23
Month and year data tags not in local timezone


mcbirse:
I've noticed what seems to be an issue with tagging of data by month/year, in that it seems to tag the data points in UTC regardless of the `tz('your/timezone')` used in the CQ query.

Here's an example, where you see the "month" tag of my data is changing month at 10am, since my timezone is +10 (at the moment) 😞 
![image](https://user-images.githubusercontent.com/108725631/191881558-00e314f9-90c9-4e18-8305-0b5f836470a6.png)

You can see the issue if you query the InfluxDB directly.

Three query examples below:

1. First showing query using time format in my timezone (for start of Sep, note month tag shows "Aug")
2. Second query is the same, but used time format in UTC and added tz() just to ensure same data is output
3. Third query shows data returned for time from 9:55am to 10:05 am - you can see this is where the month tag changes over to "Sep"
```sql
InfluxDB shell version: 1.8.10
> SELECT * FROM autogen.http WHERE time >= '2022-09-01T00:00:00+10:00' AND time <= '2022-09-01T00:05:00+10:00'
name: http
time                from_grid          from_pw            home               host month percentage         solar              to_grid            to_pw url year
----                ---------          -------            ----               ---- ----- ----------         -----              -------            ----- --- ----
1661954400000000000 7.636363636363637  316.3636363636364  318.34090909090907      Aug   56.326253953908726 -8.818181818181818 1.0909090909090908 0         2022
1661954460000000000 3.909090909090909  317.27272727272725 313.6818181818182       Aug   56.281066425666516 -8.909090909090908 7.909090909090909  0         2022
1661954520000000000 8.363636363636363  302.72727272727275 299.6818181818182       Aug   56.23587889742431  -8.727272727272727 2.272727272727273  0         2022
1661954580000000000 14.090909090909092 294.54545454545456 296.8863636363636       Aug   56.19069136918211  -8.818181818181818 1.9090909090909092 0         2022
1661954640000000000 9                  298.1818181818182  303.77272727272725      Aug   56.1455038409399   -8.636363636363637 2.5454545454545454 0         2022
1661954700000000000 9.818181818181818  300                299.52272727272725      Aug   56.09278505799066  -8.636363636363637 0.6363636363636364 0         2022
> SELECT * FROM autogen.http WHERE time >= '2022-08-31T14:00:00Z' AND time <= '2022-08-31T14:05:00Z' tz('Australia/Sydney')
name: http
[...]

PR #81, opened 2022-09-23
Add timezone to telegraf so tags are localized

mcbirse:
Add timezone to `telegraf.conf` so month/year tags are localized as per issue #80


jasonacox:
Thanks @mcbirse - a great find!  This may create some issues with `upgrade.sh` due to tz updates, but I'll see if we can mitigate that.  This is clearly needed to get the right `month` and `year` tags for the local timezone.

jasonacox:
Actually, we already addressed this in `upgrade.sh` by having it toggle timezone back to default using `tz.sh` before `git pull` so no change needed there. :)

PR #86, opened 2022-09-27
Update dashboard.json

youzer-name:
Switch Energy usage panel to time series. Enable 'centered zero' option to keep zero-points vertically centered.

Add grid_down to energy usage panel.

Remove time zone option from several panels where it has no effect when time zone is set in Grafana preferences and "use browser time" is set in dashboard settings

changed 'group by time(5m)' in queries in Energy Usage panel to 'group by time($__interval)'. It was causing fluctuations to be smoothed over. The graphs looked prettier, but the actual changes in the data were being lost.

### DIFF
--- a/dashboard.json
+++ b/dashboard.json
@@ -1,4 +1,79 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_INFLUXDB",
+      "label": "InfluxDB",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "influxdb",
+      "pluginName": "InfluxDB"
+    },
+    {
+      "name": "DS_SUN_AND MOON",
+      "label": "Sun and Moon",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "fetzerch-sunandmoon-datasource",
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks for this @youzer-name  - the json export seems to be tied to your instance.  I'm not able to get it to work. Can you export using "for sharing externally"?

image

youzer-name:
Done. Hopefully this one works.

jasonacox:
That worked!  I'm going to merge this.

jasonacox:
I noticed that the "Sun altitude" now shows up in the legend (as a max) and tool tips.  It was hidden on the older version since it doesn't add meaningful data.  We could also remove "feels like" as that likely doesn't add anything either. Thoughts?

image

jasonacox:
I went ahead and made the two minor changes and also updated temp units to "number" instead of percentage.  I also created a dashboards folder to declutter things a bit. :) 

image

youzer-name:
I'm not sure why the legend and tool tip changed, but good catch.

PR #99, opened 2022-10-13
Add tool to retrieve history data from Tesla cloud

mcbirse:
Add tool to retrieve history data from Tesla cloud as per #12 

@jasonacox - Usage instructions added to the tools README, however just thinking may want to consider moving each tools instructions to their own README perhaps? Then the main README could simply be a summary/description of available tools, with links to the specific tool. Please feel free to modify/change as required if you think this would be a good idea.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -6,6 +6,287 @@ This directory contains a list of tips, tricks and tools discovered or developed
 
 If you have a great tool or trick that you think the community would enjoy, please submit an [issue](https://github.com/jasonacox/Powerwall-Dashboard/issues) or [pull request](https://github.com/jasonacox/Powerwall-Dashboard/pulls) to get it added here.
 
+## tesla-history.py
+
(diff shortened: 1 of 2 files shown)


jasonacox:
Great work @mcbirse !!!  Thank you! This is a great addition to the project.  I agree with you on the separation of the tools.  I'll work on that.

jasonacox:
Hi @mcbirse - I moved the tools into their own folders with an index list in the [/tools](https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools) folder.  Take a look and let me know if this makes sense or if we should adjust.

mcbirse:
Hi @jasonacox - Looks great and makes sense! I think this should be much more manageable going forward if the list of tools continues to expand.

jgleigh:
I did have to move the .conf and .auth files into the new folder so the script could find them again.

jasonacox:
Ah!  That's a great point @jgleigh  - For anyone who set it up prior to me moving the tools, you will need to move your setup files into the new folder (or run the `python3 tesla-history.py --login` again).

Issue #104, opened 2022-10-15
hard coded upgrade.sh url


BuongiornoTexas:
@jasonacox, this issue may just be my inexperience with git, but I can't figure out how to test the upgrade.sh script without changing the hard coded path to the master repo, and even then, I have no idea how to make this work from a branch I'm testing (i.e. how can I force the upgrade to pull updated files from the branch I'm testing rather than the master?). 

This is in relation to updates from #96. If there is no obvious short term fix, I will raise the PR without testing upgrade.sh and will highlight that in the commit notes - there are relatively few changes and it should be straightforward to review these. 

On a related note - the upgrade.sh script runs a `git stash` before doing a rebase pull, but never refers to the stashed elements again. Is this deliberate?

jasonacox:
Great questions @BuongiornoTexas .  The intent of `upgrade.sh` was less for developers and more for end users.  In full transparency, I'm not fond of using git as the means of doing upgrades for general users. But if I pivot to having build artifacts (releases) manage state, it would likely create other complications.  But as to your points:

* I think we could update `upgrade.sh` to accommodate branching and variable repository locations, perhaps through environmental settings for developers?  What do you think?
* Thanks for your PR!
* I use `git stash` to record changes the user may have made to tracked files like `powerwall.yml` before I blow them away with `git pull --rebase`.  It means that the user can manually recover those changes `git stash pop` or just see what they are  with `git stash list`.  I debated over adding a `pop` after the rebase, but I'm worried that user changes may break the install since the tracked files should not be changed if we are doing clean upgrades.  I'm open to ideas here, but as I mention, the git method is both a blessing and a curse depending on persona of the user.

BuongiornoTexas:
@jasonacox - Your response makes sense. In answer to your own question, I think as long as changes to `upgrade.sh` are infrequent and you are willing to live with the pain of final code review, I'd stick with what you have. From my (limited) dig into, it looks as though making git behave in the way we would like for development may need a great deal of work and deliver fairly limited benefit.

And now I'll get back to trying reconcile the various energy measurements from my systems and meters - which is looking to be a long road.

jasonacox-sam:
This asked about testing `upgrade.sh` from a branch instead of the hard-coded master URL. The discussion concluded that the script is deliberately end-user-oriented rather than developer-oriented, and the maintainer chose to stick with the simple git-based approach as long as changes to `upgrade.sh` stay infrequent.

[...]

PR #105, opened 2022-10-16
Closes #96, Minor QoL enhancements

### LINKED ISSUE #96: Upgrade QoL isssues/enhancements
Hi @jasonacox,

I've just upgraded to the latest version via `upgrade.sh` and hit a couple of issues on the way - most of them due to modifying my own configuration and doing some preliminary development work. Based on these, I came up with a couple of very low priority quality of life enhancements. If you see value in any of these, I'm happy to work up PRs to gradually address them after I've got the first pass cumulative kWh PR in place.

- Could we add implement a user generated addition to `telegraf.conf` in setup and upgrade scripts - maybe appended from a `telegraf.local` file, which I think works OK with TOML? This would be helpful for maintaining personal customisations such as my three phase power monitoring, and may also be useful when developing new functionality.

- I was hit with my rootless docker and grafana permissions issue again. It would be good if we could customise `powerwall.yml` to replace the default user for grafana (maybe by environment variable)? This problem is mostly mine at the moment, but it'd be nice to sort out.

- Minor issue: Should there be a `dashboard.json` in both the root folder and in the dashboards folder? The one in the root folder has the latest changes to the temperature and sun and moon.

- **Closed in latest update**. Very minor issue: I got hit with a repeated UID error when importing the new dashboard json while keep the old dashboard in place for comparison. It may be worth changing the UID when new versions of the dashboard are released. 

- **Moved to new issue**. On dashboards, I haven't had time to figure out library panels in detail, but maybe they can be used to provide clean up updates to small elements of the dashboard, and maybe allow easier roll-your-own dashboards?

- **On to do list already** I think the influxdb.sql import raised errors about CQs already existing - maybe these should be dropped before re-adding?

BuongiornoTexas:
@jasonacox - I'm still learning my way around this - let me know if I have made any mistakes in making the PR. (just found my first error - I didn't realise I could include comments with the PR - I've copied my close out comments from #96 -sorry for the double up!).

PR implements the following:

- Key change: adds `compose-dash.sh` utility script to run docker compose, as the command line becomes unwieldy with the updates implemented in this PR. 
- Adds local `compose.env` file which provides environment variables to docker compose. Currently only implements GRAFANAUSER. 
- Overrides the telegraf command in docker compose to read additional telegraf `.conf` files from a config directory. Maps `telegraf.local` to `local.conf` in this directory. Can add non-standard measurements and/or use for local development testing without modifying the main development tree.

Setup/upgrade will silently create the local  `compose.env` or `telegraf.local` files if they do not exist, and these default settings should work for most users. Setup/upgrade will not modify these files if they exist already.

CAUTION! I've tested `setup.sh`, but don't know how to test `upgrade.sh` due to #104.


jasonacox:
Thanks @BuongiornoTexas !  Great job on these updates.  I tested an `upgrade.sh` and it worked perfectly. I just ran a fresh install and it also worked as expected. 

Thanks for the QoL updates! 🙏

jasonacox:
I discovered one glitch with this. Older versions  of docker-compose (< 1.28) do not include the `--env-file` option.  That will result in a error for setup or upgrade.  

I'm not sure if there is an easy way to detect this condition.  The alternative is to move away from the `--env-file` switch and stuff the compose.env environmental variables on the command line:

```bash
export $(grep -v '^#' compose.env | xargs) && docker-compose -f powerwall.yml up -d
```

I'll test that and update it unless you can find a more elegant solution.

BuongiornoTexas:
Ah - that was unexpected and annoying!

There are a couple of possibilities - I went with `compose.env` for maximum user transparency. However, as not many people will use it and it shouldn't be modified often, maybe we can use the default environment file for docker compose? This is `.env` and has been available since 1.8. If we modified the file name in 4 places and removed `--env-file compose.env` from `compose-dash.sh`, this might be enough to fix it.

```
echo "Running Docker Compose..."  
    # Build Docker (v1)
    docker-compose -f powerwall.yml $1 $2
else
    if docker compose version > /dev/null 2>&1; then
        # Build Docker (v2)
        docker compose -f powerwall.yml $1 $2
    else
        echo "ERROR: docker-compose/docker compose is not available or not runnning."
        echo "This script requires docker-compose or docker compose."
        echo "Please install and try again."
        exit 1
    fi
fi
```
The only wrinkle is that this file might need to be in the CWD for older versions - I think it is in the correct location for version 1.23+. See https://docs.docker.com/compose/environment-variables/ and https://github.com/docker/compose/blob/release/CHANGELOG.md (I may be digging too far back on this!).

The other option is pretty well a variation on your solution - make the same changes to `dash-compose.sh` as above, and source (`.`) the environment file to load the environment variables before running docker-compose, which I think should work, is pretty clean, and will work with any docker-compose version that supports environment variable substitution. 

 ```
# load enviroment variables into current shell.
. compose.env

echo "Running Docker Compose..."  
    # Build Docker (v1)
    docker-compose -f powerwall.yml $1 $2
else
    if docker compose version > /dev/null 2>&1; then
        # Build Docker (v2)
        docker compose -f powerwall.yml $1 $2
    else
        echo "ERROR: docker-compose/docker compose is not available or not runnning."
        echo "This script requires docker-compose or docker compose."
        echo "Please install and try again."
        exit 1
    fi
fi
```

**Comment by jasonacox** [comment]
Thanks @BuongiornoTexas - I think the `.env` may still not work for some folks (there are several OS variants and version out there including some Synology NAS based installations).  I have multiple set up so I can test, including a Ubuntu 18.04.6 LTS image set up and it is running docker-compose 1.17.1 by default.  Also, the "dot" sourcing method does not work on that Ubuntu host:

```
Running Docker Compose...
WARNING: The GRAFANAUSER variable is not set. Defaulting to a blank string.
pypowerwall is up-to-date
influxdb is up-to-date
telegraf is up-to-date
weather411 is up-to-date
Recreating grafana ... 
Recreating grafana ... done
```

It seem that Ubuntu doesn't allow auto-export even on source (it just creates a separate shell).  I was able to get it to work using this:

```bash
set -a # automatically export all variables
. compose.env
set +a
```

This still seems more elegant than my hacky env injecting method. 😁 

Minor thing but I noticed another issue in the new `setup.sh` and `upgrade.sh` and had to change all the calls to `compose-dash.sh` to be explicit calls instead of sourcing (dot) calls or it would not hand off execution to the external script (I had discovered this with `tz.sh` in the past): https://github.com/jasonacox/Powerwall-Dashboard/commit/49eccc67a5ce98c1fec233ff644f88fc73fb2be6

The challenge with this project is that you never know what type of OS or bespoke platforms like Synology NAS that may be running this.  It makes testing fun. 😉

jasonacox:
@BuongiornoTexas I updated the `compose-dash.sh` script to reflect the above. I also added a usage message if no params are provided.  And I made sure you got some credit in the comments. 😉  See https://github.com/jasonacox/Powerwall-Dashboard/commit/d725b240582f0962e9faafa985cc1f85eae6bae4.

BuongiornoTexas:
Huh! Well, the list of things I did not expect to happen today now has "Ubuntu as an edge case" at the top of it.

Nice work on putting together a one size fits all work around, and thanks for the credit.

PR #108, opened 2022-10-20
Fix issue retrieving site data of inactive sites

mcbirse:
- Fixed potential issue retrieving site data of inactive sites as per reports in #12
- Confirmed home power usage calculations are correct during VPP events (removed TODO comment)


jasonacox:
Thanks @mcbirse !  This tool is brilliant.

PR #114, opened 2022-10-27
Eliminate horizontal scrollbars on iPhone.

cwagz:
Problem: When trying to scroll through the dashboard on an iPhone, the iframe will scroll horizontally which is not desirable. This seems to eliminate the horizontal scroll bar on iPhone Safari browser for the animation iframe. It also centers the text above the grid and home icons.

I tested with my iPhone and the Chromium Edge browser on my Windows 11 PC. I do not see any negative effects. I am no longer able to scroll the iframe horizontally with safari on my iphone which was the desired result. I also notice the text is now correctly centered above the icons when viewed from Edge. It renders in Grafana like it does when going to port 8675 from the browser. I am on firmware 22.26.2 8cd8cac4 now. It seemed like the horizontal scrollbars showed up after the "foxtrot" update. Maybe this will make the iframe more tolerant to future changes from Tesla?

### DIFF
--- a/dashboard.json
+++ b/dashboard.json
@@ -962,7 +962,7 @@
       "id": 13,
       "links": [],
       "options": {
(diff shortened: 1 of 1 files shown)


cwagz:
Please let me know if I am breaking any etiquette rules or best practices with the way I did this. It is all very new to me. Thanks!

jasonacox:
Thanks @cwagz  - good find!  Thanks for submitting this PR.  Great job! 😁

PR #117, opened 2022-10-30
Updated animation and yearly image

cwagz:
Shows current dashboard with weather and usage overlay. Let me know if you would like a Last 24-hour timescale instead of a complete day. I was trying to get something that would show some clouds.


jasonacox:
Thanks @cwagz !

PR #123, opened 2022-11-08
Fix bug with error output when dateutil is missing

mcbirse:
Minor update to fix a bug with error output when the python-dateutil module is missing. Calls "sys.exit" before the sys module has been imported - whoops. 😯


jasonacox:
Thanks @mcbirse - good catch.

PR #124, opened 2022-11-11
Add tool to fix incorrect month tags of InfluxDB

mcbirse:
Hi @jasonacox 

Adding this script to the tools repository which hopefully should be useful for those affected by issue #80 and wanting to correct the data.

This python script can be used to fix incorrect month tags of InfluxDB where the month was based on UTC instead of the local timezone. My data is fixed now. 😊


jasonacox:
Brilliant!!!  Thanks @mcbirse ! 👍

Issue #149, opened 2023-01-08
Alternate Weather Source - Your Own Weather Station


BJReplay:
### Discussed in https://github.com/jasonacox/Powerwall-Dashboard/discussions/145

<div type='discussions-op-text'>

<sup>Originally posted by **BJReplay** December 26, 2022</sup>
I have a Pantech branded Home Weather Station (of course I do - I'm a gadget geek) - Ecowitt (aka Fine Offset) are the OEM for a bunch of home weather stations, including Pantech, and probably most commonly known in the US is Ambient Weather.

I've just discovered (since my internal display died last night) that they have an API: https://doc.ecowitt.net/web/#/apiv3en?page_id=19

I knew that it was possible to intercept the output having read up on weewx interceptors, but now that they have an API, it seems feasible to provide an alternative weather input to Weather411 for the dashboard.

Here's a list of weatherstations from another site that talks about accessing their data utilising the wire protocol - but I suspect this is now superseded given there is an API: https://www.openhab.org/addons/bindings/fineoffsetweatherstation/

First steps will be to get the API working in Postman, and then to figure out how to get it up and running so that it can be included into grafana - I would welcome any suggestions!

</div>

Creating Issue to reference for pull request.

BJReplay:
Closing issue - Closed with v2.7.0

PR #150, opened 2023-01-08
First Release of Ecowitt Weather Server

BJReplay:
This pull request completes most of the work to provide a local weather source based on #149 but doesn't close the issue:

- Dashboards still need to be modified to use the data
- The local weather server needs to be included into the docker stack; this can be done by including it in powerwall.yml - for example:

```
    localweather:
        build: ./localweather
        image: localweather:latest
        container_name: localweather
        hostname: localweather
        restart: always
        user: "1000:1000"
        volumes:
            - type: bind
              source: ./localweather
              target: /var/lib/localweather
              read_only: true
        ports:
            - target: 8686
              published: 8686
              mode: host
        environment:
            - WEATHERCONF=/var/lib/localweather/localweather.conf
        depends_on:
            - influxdb

```

### DIFF
--- a/weather/contrib/ecowitt/README.md
+++ b/weather/contrib/ecowitt/README.md
@@ -1,0 +1,147 @@
+# localweather Server
+
+![Docker Pulls](https://img.shields.io/docker/pulls/jasonacox/localweather)
+
+This server pulls current weather data from [Ecowitt.net Weather API](https://api.ecowitt.net/api/v3/device/real_time), makes it available via local API calls and stores the data in InfluxDB for graphing.
+
+This service was built to easily add weather data graphs to the [Powerwall-Dashboard](https://github.com/jasonacox/Powerwall-Dashboard) project.
+
+Docker: docker pull [jasonacox/localweather](https://hub.docker.com/r/jasonacox/localweather)
+
+## Quick Start
+
+
+1. Create a `localweather.conf` file (`cp localweather.conf.sample localweather.conf`) and update with your specific location details:
+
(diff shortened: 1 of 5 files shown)


weather/contrib/ecowitt/README.md
```diff
@@ -0,0 +1,147 @@
+# localweather Server
+
+![Docker Pulls](https://img.shields.io/docker/pulls/jasonacox/localweather)
+
+This server pulls current weather data from [Ecowitt.net Weather API](https://api.ecowitt.net/api/v3/device/real_time), makes it available via local API calls and stores the data in InfluxDB for graphing.
+
+This service was built to easily add weather data graphs to the [Powerwall-Dashboard](https://github.com/jasonacox/Powerwall-Dashboard) project.
+
+Docker: docker pull [jasonacox/localweather](https://hub.docker.com/r/jasonacox/localweather)
```
jasonacox, line 9: Assuming others contribute similar services for other local weather stations (or services), I'm thinking we should probably rename the container to match the specific solution it supports.   What about just "ecowitt"?

jasonacox:
Thanks @BJReplay  - excellent work!  Make sure your git commit email is set up as a github email so that you get credit as a contributor (if you care).  

This is a great addition! I added an in-line comment about the container name. I'm happy to publish that container to dockerhub, but want your thoughts on the name before I do.

weather/contrib/ecowitt/README.md
BJReplay: Agree that's a good idea.

BJReplay: I'll make those changes tomorrow, and test, and assuming it all works, create another pull request.

PR #151, opened 2023-01-08
Add Sample Local Weather Dashboard

BJReplay:
Another component for #149 - Sample Dashboard


BJReplay:
Note that this uses the old Graph type rather than the new time series type, so is saved as a sample dashboard rather than a library panel (as the library panels are disabled if they use deprecated plugins).

PR #157, opened 2023-01-14
0.0.4 - Third Release

BJReplay:
- Fix to Rainfall Data
- Fix to Humidity Data
- Add Dew Point to Weather Data
- Change default name of conf file
- Correct Name of config file comment in server.py
- Fix Build Your Own instructions in README (incorrect container name)
- Remove redundant data


BJReplay:
There's a bit of reorganisation that was done to help me get a history loader up and running (pull request to follow once I think it is ready) that hides some of the fixes (some incorrectly named fields) around rainfall data being stored in influxdb as hourly instead of rain_1h which was the intent.

jasonacox:
Thanks @BJReplay !  0.0.4 pushed: https://hub.docker.com/r/jasonacox/ecowitt/tags

PR #159, opened 2023-01-16
Ecowitt Weather History Tool

BJReplay:
Shamelessly inspired by @mcbirse this pulls in history from the Ecowitt API and populates the localweather measurement.

No remove function - as localweather is so new I don't think it's needed :)


jasonacox:
Nice!  Thanks @BJReplay !

jasonacox:
It may make sense to reference this tool in the Ecowitt [README](https://github.com/jasonacox/Powerwall-Dashboard/tree/main/weather/contrib/ecowitt#ecowitt-local-weather-server)  so other Ecowitt users can find it.

PR #163, opened 2023-01-26
Change frequency panel to 3 decimal places

longzheng:
The frequency panel is a bit hard to understand because frequency is commonly fluctuating only between a very small range. 

The hover tooltip becomes a bit useless at one decimal place because it rounds the number and it's not possible to distinguish when the value changes.

image

Changing to 3 decimal places allows you to see the actual fluctutation clearer.

image


jasonacox:
Good enhancement.  Thanks @longzheng !

PR #168, opened 2023-02-01
Update README.md

BJReplay:
Indent powerwall.yml snippet so that it can be cut and pasted directly into powerwall.yml

Minor doco change once you've completed the pull requests and branch merge for the 230V work.


jasonacox:
Thanks @BJReplay  - Good idea!

PR #182, opened 2023-02-19
Add alerts CQ and revert telegraf changes, update run-once and dashboard

mcbirse:
Hi @jasonacox - Here are the changes to use a CQ for the alert data, as below:

- Reverted "telegraf.conf" changes, as they are not required (also fixed some whitespace/tab inconsistencies)
- Added new continuous query "cq_alerts" to downsample data from raw to the alerts rp in per minute intervals
- Updated "dropcq.sql" script to include all retention policies (in case used in future - not required for this upgrade)
- Updated dashboard alert field names to include "max_" prefix, and added rename by regex transform to display original field names in dashboard graphs
- Updated "run-once-2.8.3.sql" script so the last 3d of data is copied to the new alerts retention policy during upgrade

Just a note on the new alerts continuous query and why I have added a `RESAMPLE FOR 2m` clause. During testing I observed something interesting happen...

We (conveniently?) had a pretty severe storm come through south of Sydney last night, while I had my test system with these changes running. I manually switched my Powerwall to Off-Grid for safety reasons, which you can see in the alerts below.

What I noticed however was, the alert "SwitchingBridgeIrrational" that occurred at 18:46:55 (5 seconds before the CQ would run), did NOT get downsampled into the per minute data of the alerts rp (i.e. it would never have been seen/noticed/logged etc., unless you manually scrutinised the raw data).

image

So it appears it could happen occasionally (perhaps due to system load or timing issues), a CQ running e.g. per minute may miss a data point logged just before it's execution time.

So by adding a `RESAMPLE FOR 2m` clause to the CQ, this means when the CQ runs every minute (as per the `GROUP BY time(1m)`) on each execution it will resample _the last 2m of data_ - this should mitigate against the possibility of missing an alert.

I did some testing of this by manually adding data points with a timestamp in the past minute after a CQ would have already executed, and it was then downsampled to the per minute rp in the next run of the CQ.

For alert data I believe this could be important to ensure we capture any alerts that occur.

Reviewing the other CQ's however... I don't think we need to be concerned about this potential issue - the only other CQ that may benefit from a similar change could be "cq_grid", however grid status would be logged as an alert anyway so it probably isn't necessary.


jasonacox:
Brilliant @mcbirse !  You're amazing. Thank you!  🙏 

The RESAMPLE  addition makes a lot of sense and I agree with you, with alerts we don't want to miss the signal. That was originally the reason I was willing to spend the space for raw storage, but ironically, as you discovered, it would have resulted in short-events being missed on the graph due to being too small to render.

PR #185, opened 2023-02-20
Support customised powerwall.yml through upgrade.

BJReplay:
Supports a customised powerwall.yml Docker Compose file through the upgrade by preserving it before the upgrade by copying it to a .restore copy, and restoring it after the git pull before restarting the containers.

Note that I haven't actually been able to test this, given that the upgrade.sh script runs itself from the latest version from your repo; I'm pretty confident, though; what I'm going to do is create this pull request, then edit my script again so that it pulls from my fork (and thus gets my new script) to test that it works - I'll comment back here again to confirm it works.

### DIFF
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -72,6 +72,17 @@ else
     exit 
 fi
 
+# Preserve Customised powerwall.yml?
+read -r -p "Do you want to preserve a customised powerwall.yml (only for specialised users)? [y/N] " response
+if [[ "$response" =~ ^([yY][eE][sS]|[yY])$ ]]
+then
+    echo "preseving powerwall.yml"
+    cp "powerwall.yml" "powerwall.yml.restore"
+else
+    echo "Replacing powerwall.yml with standard as part of upgrade"
+fi
+
+
 # Remember Timezome and Reset to Default
 echo "Resetting Timezone to Default..."
 DEFAULT="America/Los_Angeles"
@@ -139,6 +150,12 @@ if ! grep -q "TZ=" pypowerwall.env; then
     echo "Adding..."
     echo "TZ=America/Los_Angeles" >> pypowerwall.env
 fi
+
+if [ -f powerwall.yml.preserve ]; then
+    echo "restoring preserved powerwall.yml"
+    cp "powerwall.yml.restore" "powerwall.yml"
+fi
+
 
 # Make sure stack is running
 echo ""
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -63,6 +63,15 @@ if [ ! -f ${ENV_FILE} ]; then
 fi
 
 # Verify Upgrade
+echo "If you have customised your powerwall.yml file, you can preseve it through the upgrade"
+echo "This feature is only for advanced users who know what they are doing and have reviewed"
+echo "any changes in the powerwall.yml file before starting this upgrade."
+echo "If you thing you might want to do this, but have not reviewed your file, cancel the"
+echo "upgrade at the next step, and review your powerwall.yml file for any changes first."
+echo ""
+echo "This feature can be ignore by most users."
+echo ""
+echo ""
 read -r -p "Upgrade - Proceed? [y/N] " response
 if [[ "$response" =~ ^([yY][eE][sS]|[yY])$ ]]
 then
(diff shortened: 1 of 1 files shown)


BJReplay:
Yep, it worked.

It does expose a bug, though - I think the script exposes a bug where it assumes that Weather411 always exists (or maybe, because I used to run it, the logic to check still thinks it is there) - the failure was in the section that reports that it is about to `Delete and Upgrade weather411 to Latest`

mcbirse:
@BJReplay - I was wondering if there was an alternative method of implementing a customised Docker Compose configuration, while keeping the default powerwall.yml intact (eliminating the need to save and restore the config file during upgrades).

I don't have much experience with Docker, however just found the following which sounds promising: https://docs.docker.com/compose/extends/

Some more investigation would be required, but it appears it would be possible to for example have a `powerwall.override.yml` compose file whereby one could implement their customisations/extension to the base `powerwall.yml` configuration?

Ensuring this override file was not tracked would mean it would not be affected during upgrades (a `powerwall.override.yml.sample` file could be added as guide which a user could copy and customise).

I've only looked at this option for a few minutes though, so I may have misunderstood how it works or your requirements.

jasonacox:
Hi @BJReplay - thanks for the PR!  A few things as I review this (ok, a small book 😂 ):

We have tried really hard to make all custom settings external to `powerwall.yml` so that upgrades to `powerwall.yml` wouldn't break user local settings. There have been several versions that made changes to `powerwall.yml` in the past to add enhancements.  My question to you: Is there a way you can get your customization without changing powerwall.yml?   And, if we accommodate this customization, are we setting ourselves up for a flood of issue tickets to troubleshoot edge cause custom problems?

I believe that most of the members of our community just want the "out of the box" dashboard they can play with and tweak. I propose that they should not be preserving their `powerwall.yml` (obviously I don't :), only those who know how to fix their system if/when the upgrade breaks.  I know that you are introducing a conditional prompt so that someone can choose to preserve the file or not, but I would prefer that we make it less obvious and make the option available via a command line switch. I don't want to tempt anyone to select it casually.  Perhaps `upgrade.sh --preserve` and then it presents a stern "WARNING: By preserving your powerwall.yml, upgraded features or setting may not work or could break your setup"?  I'm just worried that if we make it too easy, it will generate 10x number of issues that we have to address helping our users fix their system after an "upgrade". 😛 

On Weather411: The default install will always load the weather411 container. It will run and if the `weather/weather411.conf` file is missing (not set up), it will go into idle mode, but still remain as a running container.  The reason is that if the user later wants to add weather, they run the `./weather.sh` script and it creates the config file which activates the weathe411 container. This also makes upgrades straightforward.  In your case, since you modified `powerwall.yml`, you likely removed weather411 which upgrade.sh always expect to be there.  It also means that the weather.sh script won't work for you either. Of course, you are savvy enough to know how to fix that. But many in the community are just learning, which is exciting and want to encourage, but I just don't want to make more work for myself troubleshooting problems we can avoid with standardized automation. 

Just so I am not misunderstood, I want to be clear that this project is an enabler for people to experiment and customize all they want to. But tools like `upgrade.sh` are not for the experts.  It is aimed at the masses. That is by design so that key contributors like yourself and I, don't end up spending many hours troubleshooting edge cases and can actually work on new features instead. 😅  

If that makes sense, I'm happy to accept a merge that changes this feature to a command line option. Thoughts?

BJReplay:
Thanks Gents!

@jasonacox - I agree, a command line option is a much better approach.

I take your point about running the Weather411 container.

@mcbirse - let me look at the extends documentation you pointed out.

Once I've had a chance to look at both, I'll either submit an updated pull request for a command line (or perhaps close this without needing any change).

While I've got your attention, I ran into a minor related issue - I customised the backup.sh script (changed `DASHBOARD="/home/user/Powerwall-Dashboard"    # Location of Dashboard to backup` from /home/user to reflect my install), and (of course) the backup.sh script was overwritten as part of the upgrade.

I wonder if that means that the backup.sh script should be renamed to backup.sh.sample, or the instructions should suggest using it as a sample, and copying it to a different name so it doesn't get overwritten.

BJReplay:
OK, the extend works.  I'll close this pull request, and open a pull request for a tested change to compose-dash.sh instead :)

BJReplay:
Closing. Not Required

mcbirse:
> I wonder if that means that the backup.sh script should be renamed to backup.sh.sample, or the instructions should suggest using it as a sample, and copying it to a different name so it doesn't get overwritten.

@BJReplay - This sounds like a good idea. I had the same issue and copied the script, then customised it for my system, added to cron, etc. If it were named "backup.sh.sample" it would be more obvious to do this. Maybe add that change in your next PR as well? 😊

PR #186, opened 2023-02-20
Support powerwall.extend.yml & backup.sh.sample

BJReplay:
Apologies for six different commits.  The first four are two commits and two more undoing them - the changes to update.sh.

The next two are the meat on the bones.  The changes to compose-dash.sh look for the existence of powerwall.extend.yml, and if it exists, initialise a variable with -f powerwall.extend.yml to be included in the docker compose or docker-compose commands (whatever they are) so that the docker compose command works on both powerwall.yml and powerwall.extend.yml.

I used this to shut down my customised powerwall.yml, bring up the standard one (with Weather411 added back in), shut it down, bring up one with Ecowitt added back in (by creating powerwall.extend.yml).

The other change here is to the snippet in the instructions for the Ecowitt weather service to add in the version and services tags that allow this yaml to be merged.

Associated change to the readme noting this.

Final change here is a rename of backup.sh to backup.sh.sample, and associated change to the readme: I got bitten by the upgrade today when my customised backup.sh got overwritten by the upgrade.sh and a second commit with improved instructions.


jasonacox:
Excellent!  Thanks @BJReplay ! Good enhancement that creates a clean and supported path to add custom docker additions or tools. 🙏 And, brilliant call on `backup.sh`.  

Thanks @mcbirse for spotting the docker compose extends path.

jasonacox:
Released as v2.8.4 - Custom Docker Compose Support

Thanks again @BJReplay !!

PR #197, opened 2023-02-27
Upgrade InfluxDB Client

BJReplay:
As per your changes for Weather411.

Duplicated.  Not (yet) tested.

I'll do a docker build and come back and comment on this pull request once I've confirmed that it works 😉

### DIFF
--- a/weather/contrib/ecowitt/server.py
+++ b/weather/contrib/ecowitt/server.py
@@ -47,6 +47,13 @@
         PORT = 8086
         DB = powerwall
         FIELD = ecowitt
+        # Auth - Leave blank if not used
+        USERNAME =
+        PASSWORD =
+        # Auth - Influx 2.x - Leave blank if not used
+        TOKEN =
+        ORG =
+        URL =
 
     ENVIRONMENTAL:
         LOCALWEATHERCONF = "Path to localweather.conf file"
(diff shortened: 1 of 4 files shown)


jasonacox:
> I had to add back in the removed datetime.datetime 

To make that work (datetime instead of datetime.datetime), make sure you change the import to this:

```python
from datetime import datetime
```

weather/contrib/ecowitt/server.py
```diff
@@ -275,11 +289,19 @@ def fetchWeather():
...
+                                    database=IDB)
+                            else:
+                                # Influx 2.x
+                                client = InfluxDBClient(
+                                    url=IURL,
+                                    token=ITOKEN,
+                                    org=IORG)
                             output = [{}]
                             output[0]["measurement"] = IFIELD
                             output[0]["time"] = int(currentts)
```
jasonacox, line 307: I had to change this line - I don't know if this is an issue with the new client or just best practice:

```python
# for weathe411
# was: output[0]["time"] = weather["dt"]
# now:
output[0]["time"] = datetime.utcfromtimestamp(weather["dt"])

# yours?
output[0]["time"] = datetime.utcfromtimestamp(int(currentts))
```

BJReplay: I had a look at that change in Weather411, but I had a difference between ecowitt and weather411 in this section in the original version - Weather411 has a date that can be parsed, whereas ecowitt doesn't, which is why my version just passes in a truncated (int) timestamp, so I didn't think it needed to be changed.

I'll have a look for a missing from datetime import datetime, but I would have thought that would have shown up in a diff - perhaps I wasn't paying attention :)

But the bigger issue is the missing api_client.

Anyway something for another day.

PR #204, opened 2023-03-05
Add support for Solar Only Tesla energy sites to tesla-history script

mcbirse:
Added updated version of tesla-history.py script that (hopefully) supports retrieval of historical data from Solar Only Tesla energy sites, per #183.

Currently located in [Solar Only dashboard placeholder](https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools/solar-only) for testing only.

- Now retrieves list of sites for all types of Tesla energy products (battery & solar)
- Conditional history data retrieval based on site type (i.e. charge percentage & grid status only relevant for Powerwall sites)
- Testing required (debug mode) to determine what power history data is returned for solar only sites and which fields should be written to InfluxDB


jasonacox:
Thanks @mcbirse ! 🙏

PR #226, opened 2023-03-27
Fix timezone for solar only sites using an offset

mcbirse:
Fix timezone for solar only sites using an offset as per observations by @youzer-name in #183 

Due to incorrect behaviour of Tesla cloud history data when the timezone is reported as an offset, the offset will now be replaced with the configured InfluxDB timezone instead.

### DIFF
--- a/tools/solar-only/tesla-history/tesla-history.py
+++ b/tools/solar-only/tesla-history/tesla-history.py
@@ -224,6 +224,7 @@ eventdata = []
 reservedata = []
 sitetz = None
 tzname = None
+tzoffset = False
 power = None
 soe = None
 backup = None
@@ -399,7 +400,7 @@ def get_power_history(start, end):
 
     Adds data points to 'powerdata' in InfluxDB Line Protocol format with tag source='cloud'
     """
-    global sitetz, tzname, dayloaded, power, soe
+    global sitetz, tzname, tzoffset, dayloaded, power, soe
 
     print(f"Retrieving data for gap: [{start.astimezone(influxtz)}] - [{end.astimezone(influxtz)}] ({str(end - start)}s)")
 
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks @mcbirse !  It's a bit like coding in the dark? :)

mcbirse:
> It's a bit like coding in the dark?

Very much so! ☹️

youzer-name:
@mcbirse 

I modified the script with these changes and loaded the data for the 2021 time zone changes.

For March 2021, everything during standard time was off by one hour, and it started being correct the Monday after the switch to daylight saving.

For November 2021, everything during DST was correct, and it started being off by one hour on the Monday after the switch to standard time.

Setting aside the weirdness with the day of the change, it looks like it isn't using the correct offset during standard time (when I am currently pulling the data during daylight saving time).

mcbirse:
@youzer-name - Thanks. I haven't really had a chance to look into this further yet.

Can you confirm this section of code is running for you (from line 487). Adding `print(timestamp)` before and after timestamp is assigned would show that as well as be helpful to see if it is being changed.

```py
                # Check if solar only site timezone is using an offset and replace with InfluxDB timezone
                if isinstance(site, SolarPanel) and tzoffset:
                    print(timestamp)
                    timestamp = isoparse(d['timestamp']).replace(tzinfo=influxtz).astimezone(utctz)
                    print(timestamp)
```

I thought this would change the UTC-4 offset to UTC-5 offset during non-DST periods, and that was going to fix it - obviously not. It is so hard when not able to test this myself.

youzer-name:
Ugh... operator error. I was editing a copy of the script in one folder and then running a different copy from a sub folder, so my last debug output was from an old version of the script! Now that I ran the correct script, the data seems to line up both before and after the daylight saving change, but as expected it is incorrect on the day of the change just like it is in the Tesla app.

So your latest version works for my solar-only with an offset.

mcbirse:
> So your latest version works for my solar-only with an offset.

Phew! 😅 

I honestly thought the last change would in fact fix it, but then when you posted above that there was still a problem I was scratching my head looking at the code and re-testing things, completely miffed as to why it wasn't working!

Hence why I started to think that bit of code was not being run at all for some reason.... 👀 Lucky I asked before randomly/blindly starting to make more changes.

All good, glad it finally works!! I don't think we can do much about the change-over days being wrong however.

Also, I think I will leave this specific to "solar only" sites. I considered making this change for any site type that uses an offset for the timezone (i.e. Powerwall sites) but I have not yet observed any Powerwall sites like this - they all seem to use a timezone name.

If it does occur, I could make further changes, however the fix would need to be applied for the "soe" and "backup" event data as well (not relevant to solar only sites) which makes it a bit more complicated.

PR #232, opened 2023-04-01
Various updates

ThePnuts:
Updated instructions to include the plugin for Current State panel and added it to the dashboard

image

Added the scaling Power Flow iFrame


tools/influxdb2/README.md
```diff
@@ -82,7 +82,8 @@ This includes optionals files and instructions to help setup InfluxDB 2.x (inste
     * Enter the Organization you created
     * Default Bucket: tesla_energy
   * Add plugins in configuration
-    * Search for and isntall: Sun and Moon (By fetzerch)
+    * Search for and install: Sun and Moon (By fetzerch)
+    * Search for and install: Boom Table (By yesoreyeram)
```
jasonacox, line 86: Note only: Most people will likely be able to use the Environmental variables to have Grafana install those plugins itself. I know your setup makes that difficult (restricted internet access?) but we could add a note to set the env vars when you start the Grafana container:

https://github.com/jasonacox/Powerwall-Dashboard/blob/0f8468b27468791ad882f53b5d58cebb748e184a/grafana.env.sample#L1-L9

ThePnuts: I may be over complicating it the way I did it at the time, I wasn't aware of the global variables and set it up to actually edit the config file instead so I could do this: GF_PANELS_DISABLE_SANITIZE_HTML=true as I noted in the setup.

I will look at removing this and implementing the variables as that is definitely the better way.

For GF_PATHS_PROVISIONING=/var/lib/grafana/provisions 
What is this doing exactly?

For GF_INSTALL_PLUGINS, I believe I was able to recreate the entire dashboard with only adding Sun and Moon, and Boom Table. You don't happen to know what panels you used the other plugins for do you?

I also did install influx1.8 and have my solar data going into both DB's so I could import your normal dashboard into grafana for comparison using influx1.8. The only other plugin that is erroring is the pie-chart, but its because it is using the legacy version vs the current one thats installed with grafana now

jasonacox: > For GF_PATHS_PROVISIONING=/var/lib/grafana/provisions
> What is this doing exactly?

You can provide a config to auto-provision grafana which would expedite the setup (no user input required to specify the InfluxDB details, dashboard.json, etc.). This setting is needed to do that but I never got around to building configs that would actually work.  More TODO backlog. 

> The only other plugin that is erroring is the pie-chart, but its because it is using the legacy version vs the current one thats installed with grafana now

Yes, great point! We need to clean up the legacy plugins too. As the default dashboard has evolved (thanks to this community), many of the plugins are no longer needed.  Cleaning it up will allow Grafana to startup faster. Also probably time to test latest and see if we can pin to a newer Grafana version after we test it across the most common platforms (RPi, Win, MacOS, Linux, i86/64 and arm). Another TODO... of course, any help testing or PRs appreciated. 😉

PR #233, opened 2023-04-01
Few fixes and moved to grafana variables use instead of grafana.ini

ThePnuts:
Adjusted some panels to better match official dashboard release.

Removed use of editing grafana.ini file and instead moved to using variables.
Currently plugins are now auto installed (2 identified as required) and html is allowed via variable.


jasonacox:
Thanks @ThePnuts !

jasonacox:
At one point, we added the alpha version of yesoreyeram-boomtable-panel because the default version had a bug that cause "No data" to display for readings.  Are you seeing the "No data" bug? We probably need to see if the current version finally merged the fix.

Alternatively, you can specify the boomtable version to pull via:

```
https://github.com/yesoreyeram/yesoreyeram-boomtable-panel/releases/download/v1.5.0-alpha.3/yesoreyeram-boomtable-panel-1.5.0-alpha.3.zip;yesoreyeram-boomtable-panel
```

PR #247, opened 2023-04-18
Fixed some formatting and updated auto powerwall resize

ThePnuts:
Few other minor changes I saw, such as using max instead of mean to try to keep peaks on energy use chart, etc.


jasonacox:
Thanks @ThePnuts ! 🙏

PR #259, opened 2023-04-27
Merge changes to support solar only sites

mcbirse:
Hi Jason - suggesting we merge all current changes of the tesla-history script from the solar only site testing to the main script location. I believe it is working well to pull history for both Powerwall sites and/or Solar-Only sites now.

Currently working on further changes to the script though, but will update this in the solar only testing location when ready. Some progress made in the last few evenings to support daemon mode so we can containerize the script hopefully.


jasonacox:
Merged!

Nice work @mcbirse !  I agree with you on combining these. It will be easier to maintain. Thank you! 🙏

PR #266, opened 2023-05-03
Make costs/credit per kWh and timezone to be variables in Grafana

### LINKED ISSUE #263: Move hard-coded avg cost per kWh into a variable
**Problem**
Every time there's an update to the Grafana dashboard(s) users need to go and change each query where 0.19 is used.

**Enhancement**
0.19 can be moved to variables and later used in queries making it easier to adjust after updating dashboards.

**Additional context**
Something like this can be added to the dashboard model and $avg_per_kwh can be used in queries where appropriate.
```
  "templating": {
    "list": [
      {
        "description": "Average Cost Per kWh",
        "hide": 2,
        "label": "kWh cost",
        "name": "avg_per_kwh",
        "query": "0.19",
        "skipUrlSync": false,
        "type": "constant"
      }
    ]
  }
```

emptywee:
This closes #263 

However, I do not differentiate between costs and credits here, assuming both are same. I do understand that buying electricity may include delivery costs, which you probably don't get compensated for when you export electricity, so as part of further improvements, we can set up two variables: price per kWh when you import (buy) and price per kWh when you export (sell) and use them accordingly in the queries. This is up for discussion.

Now, when you import a dashboard, you get prompted like this:

image

Second thing: there's too many dashboards in the repository, I got confused a bit, and edited only those that I kinda use or understand what they are. Hopefully it's good enough for starters.

### DIFF
--- a/dashboard.json
+++ b/dashboard.json
@@ -15,6 +15,20 @@
       "type": "datasource",
       "pluginId": "fetzerch-sunandmoon-datasource",
       "pluginName": "Sun and Moon"
+    },
+    {
+      "name": "VAR_AVG_PER_KWH",
+      "type": "constant",
+      "label": "Cost per kWh",
+      "value": "0.19",
+      "description": ""
+    },
+    {
+      "name": "VAR_TZ",
+      "type": "constant",
+      "label": "Timezone",
+      "value": "America/Los_Angeles",
+      "description": ""
     }
   ],
   "__elements": {},
@@ -614,7 +628,7 @@
           "measurement": "http",
           "orderByTime": "ASC",
           "policy": "autogen",
-          "query": "SELECT \"mean_load_instant_power\" FROM \"http\" WHERE $timeFilter tz('America/Los_Angeles')",
+          "query": "SELECT \"mean_load_instant_power\" FROM \"http\" WHERE $timeFilter tz('$tz')",
           "rawQuery": false,
           "refId": "A",
           "resultFormat": "time_series",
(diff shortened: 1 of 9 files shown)


emptywee:
Feel free to make any changes to labels/descriptions/etc.

dashboard.json
```diff
@@ -15,6 +15,20 @@
...
+    },
+    {
+      "name": "VAR_AVG_PER_KWH",
+      "type": "constant",
+      "label": "Cost per kWh",
+      "value": "0.19",
+      "description": ""
+    },
+    {
+      "name": "VAR_TZ",
```
jasonacox, line 27: Nice!  I love that you added this.

dashboard.json
```diff
@@ -15,6 +15,20 @@
...
+      "type": "constant",
+      "label": "Timezone",
+      "value": "America/Los_Angeles",
```
jasonacox, line 30: The `tz.sh` script will still update this in the `dashboard.json` file to whatever the user enters during `setup.sh` so the input should default to the users preferred timezone.  That should be fine. I still like this as an input var.

jasonacox:
Merged!

Thank you for this @emptywee !!! 🙏   This is a great feature/enhancement.

I tested these on my systems with no issues.  I'll fold this into v2.9.3

## v2.9.3 - Adjustable kWh Cost

* Dashboard Update - Make "costs/credit per kWh" and "timezone" user adjustable variables in Grafana by @emptywee in https://github.com/jasonacox/Powerwall-Dashboard/pull/266

jasonacox:
Thanks @jgleigh  - I updated with your JSON and committed as a new branch.  See https://github.com/jasonacox/Powerwall-Dashboard/commit/9bffaadd04710bd36c8eb633e0957aba80c0291c

Take a look and let me know if it looks good.

Issue #155, opened 2023-01-12
backup on a windows machine 


prussell69:
After having a disaster on my teslamate which I managed to recover, I decided it would be a good idea to "backup" my powerwall dahsboard too.
When I tried to follow your directions using git bash from my powerwall-dashboard folder, the backup.sh would only return "Must run as root".
Since I'm not very savy, I must be doing something wrong. I tried setting git bash to "run as administator" but that didn't seem to work either. Also sudo -i doesn't work. I even tried using windows terminal.
I'm sure you've probably already answered someone else, but I couldn't find anything similar.
What am I doing wrong???

Please help...

BJReplay:
sudo -s

Starts a root shell on WSL

That might be enough to then run backup.sh as root.

exit to exit from your root shell.

prussell69:
Tried sudo -s on the command line and I get:
sudo : The term 'sudo' is not recognized as the name of a cmdlet, function, script file, or operable program. Check
the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ sudo -s
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (sudo:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
If I try wsl, I only seem to get various errors, depending on what I try.
Since I'm not a Linux user, I seem to struggle with some of these suggestions.

BJReplay:
I assumed that you had installed this under WSL, but it seems you didn't, so I probably can't help, other than to suggest you try again running an elevated command prompt, and changing to your powerwall-dashboard folder to run backup.sh from there.  But even if you start bash from an elevated command prompt or powershell, I still think you need to run sudo -s to run as root within bash.

jasonacox:
Hi @prussell69 - the only reason `backup.sh` needs to run as root is to ensure it has access to grab all of the InfluxDB files that get created under the docker root user. Check to see if you have any root owned files:

```bash
ls -lR | grep root
```

If you get anything like this, it means you need to be root to backup this data:
```
drwx------ 2 root root 4096 Jan 10 16:09 1400
drwx------ 2 root root 4096 Jan 11 16:10 1402
drwx------ 2 root root 4096 Jan 12 16:09 1404
drwx------ 2 root root 4096 Jan 12 16:00 1406
ls: cannot open directory './influxdb/wal/powerwall/autogen/1144': Permission denied
```

[...]

Issue #265, opened 2023-05-02
Repo cleanup.. remove ignored files, update ignore list


YesThatAllen:
_this is an offer for me to help, I'm looking for confirmation & maybe some guidance_

**Problem**

I'm seeing a number of changes in my local git gui which could be prevented with some cleanup.

**Enhancement**

- There are some `.DS_Store` files in the repo which are ignored & show as diffs
- it looks like there are some database files in the repo which need cleanup
- adding some well placed `.keep` files can ensure folders are in place upon git checkout
- I see some changes in the backup script I'd submit in a separate PR once this is done.
 
**Additional context**

I've been running cross platform dev teams for a while. Not everyone knows about a `.gitattributes` file or `.editorconfig`.
I'm not super familiar with how to help the actionable code, but I can sure clean up a repo. 😄 

I plan on opening a PR with changes soon, I'll have some questions about how best stuff should work.

My local diff in the moment:

![2023-05-02_10-12-57](https://user-images.githubusercontent.com/2879972/235709106-af5b889e-64d7-459d-8caa-5f4864b314b3.png)

jasonacox:
Hi @YesThatAllen  - I would love a PR for this!  Thank for your help. 🙏

YesThatAllen:
Here's a start to the work.. all feedback is welcome. 

I've got a question regarding timezones.. I wonder if the diff I'm about to paste in could be handled in a separate issue regarding timezone support, but maybe I'm setting my tz to Chicago the wrong way, resulting in the diff.

Diff:

jasonacox:
The `tz.sh` script updates all those files to have your timezone and is run during `setup.sh`.  The `updated.sh` uses `tz.sh` to reset to default (yes, America/Los_Angeles) and then does a `git pull` and runs `tz.sh` again to set it back to your local timezone as stored in the local (untracked) `tz` file.

Yes, it is a hack, but the timezone localization required in all those files didn't allow for an external reference.

YesThatAllen:
Gotcha.. Once we're done with #269, there will be FAR fewer diffs showing, and in my mind, those will be easier to ignore. Addressing the "hack" is fodder for a later date/issue,

YesThatAllen:
Gotcha.. Once we're done with #269, there will be FAR fewer diffs showing, and in my mind, those will be easier to mentally ignore.

Addressing the "hack" is fodder for a later date/issue.

jaydkay:
[...]

PR #270, opened 2023-05-06
add .gitattributes file

YesThatAllen:
Prevents issues such as .sh files being borked on windows OS (ref https://github.com/jasonacox/Powerwall-Dashboard/issues/155#issuecomment-1381300040 )


YesThatAllen:
This is related to #265 but I'm putting it in a separate branch because merging this may cause some issues on various users' locals.

It's like ripping off a band-aid though.. better to address any pain & get it over with.

jasonacox:
> this may cause some issues on various users' locals.

What would be the expected scenario where that occurs?  A vast majority of our community just use `setup.sh` and `upgrade.sh` with no customization.  I tend to optimize and accommodate those to ensure no impact for upgrades. This seems sane/safe, just asking for edge cases that might impact the majority.

Thanks for your expertise! 🙏  I had no idea about these files and intend to explore them for my other projects. 😉

YesThatAllen:
Yea, sorry to have gone all doom & gloom here 😱, it shouldn't be a big deal.

The `.gitattributes` file is mostly focused on unifying line endings:
 - linux's `lf`
vs 
 - windows's `crlf` 

linux basically sees `crlf`  as two line endings, and `bash` just can't cope (that's the weirdness you saw in #265 )
So, this helps with all that.

The worst issues I've seen relate to image files & files with unicode names.
If the repo had image or other binary files, we'd need to make sure their line endings were binary.. eg `*.png  binary` so that git would know to leave the line endings alone.

What I don't know is if people will see issues with `git stash` & `git pull --rebase` when trying to retain changes.

Worst case scenario: a .sh file was edited locally on a windows computer, and had `crlf` line endings.. and now the script is going to force it to `lf` line endings.

Should be easy enough to test on a windows box, I just don't have a good test bed (with docker on windows, etc) handy. AND I also don't think you suggest edits to files which are in the repo (untracked files shouldn't be affected.


I guess the biggest issue is hitting a problem mid-upgrade script. What would happen if the script failed here [upgrade.sh#L84-L90](https://github.com/jasonacox/Powerwall-Dashboard/blob/0f150c780dae323b27d94bcd8b9f0f30f8315ae1/upgrade.sh#L84-L90) ?

YesThatAllen:
I just used `git add --renormalize .` on the repo, to see what other issues might come up. Two out of six `.flux` files had been committed as `crlf` to I ~fixed~ unified those.

And, for the record.. there's no need to add every kind of extension.. that first line says "git, you can pretty well assume files are plain text, and you can set line endings according to the current OS default." I'm doing it more for clarify.

jasonacox:
Thanks again @YesThatAllen !  I merged and tagged this as v2.9.6.

PR #283, opened 2023-05-22
Revise solar-only setup script

mcbirse:
Revised solar-only setup script as per [#183](https://github.com/jasonacox/Powerwall-Dashboard/issues/183#issuecomment-1556463759)

The script will now run tesla-history setup process by using the docker container, which removes the requirement for the local environment to have the correct python version or modules.

### DIFF
--- a/tools/solar-only/setup.sh
+++ b/tools/solar-only/setup.sh
@@ -105,19 +105,6 @@ if [ -f weather.sh ]; then
     ./weather.sh setup
 fi
 
-# Setup Tesla-History Server
-if [ -z "${TZ}" ]; then
-    TZ="${CURRENT}"
-fi
-echo ""
-echo "Tesla-History Setup"
-echo "-----------------------------------------"
-echo "Checking/installing python modules..."
-cd tesla-history
-pip install python-dateutil teslapy influxdb > /dev/null 2>&1
-python3 tesla-history.py --setup --timezone "${TZ}"
-cd ..
-
 # Build Docker in current environment
 ./compose-dash.sh up -d
 echo "-----------------------------------------"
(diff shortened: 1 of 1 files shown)


mcbirse:
Hi Jason - let us know if this change is okay. I merged already so Jot18 can test it.

Happy for other ideas or refinements to the script to make the install process more robust. I tested quickly only, but it seems to work well and the setup process flows nicely.

This is definitely a lot better than relying on python on the local machine. The setup process also does not need to check/install missing python modules either (that can be left to user if for instance they wanted to run the tesla-history script directly to import historical data - which btw will work with the same config/auth created for the docker container).

tools/solar-only/setup.sh
```diff
@@ -150,6 +137,13 @@ if [ -f weather/weather411.conf ]; then
     docker restart weather411
 fi
 
+# Setup tesla-history
+if [ -z "${TZ}" ]; then
+    TZ="${CURRENT}"
+fi
+echo "Setup tesla-history..."
+docker exec -it -w /var/lib/tesla-history tesla-history python3 tesla-history.py --setup --timezone "${TZ}"
```
jasonacox, line 145: I noticed that windows systems don't seem to run docker interactive mode (`exec -it`) very well.  I had to use `--tty` only for `verify.sh`.   I'll try this tonight on a windows system when I can.

jasonacox: Yes, as I suspected.  This happens on my Windows 11 system:

```bash
$ docker exec -it pypowerwall sh
the input device is not a TTY.  If you are using mintty, try prefixing the command with 'winpty'
```

But it does work with winpty prefix:

```bash
$ winpty docker exec -it pypowerwall sh
/app $ uname
Linux
/app $
```

jasonacox: I wish I knew a Windows guru that could help. I suspect we could just treat this with a condition to the script like this:

```bash
if type winpty > /dev/null; then
    winpty docker exec -it -w /var/lib/tesla-history tesla-history python3 tesla-history.py --setup --timezone "${TZ}"
else
    docker exec -it -w /var/lib/tesla-history tesla-history python3 tesla-history.py --setup --timezone "${TZ}"
fi
```

jasonacox:
I added the condition for winpty for Windows systems. It seems to work for me and haven't detected any side effects.  

By the way, it seems to me that WinOS "Linux" subsystem is far from "standard" and always requires custom handling. 🤷

PR #285, opened 2023-05-27
Add InfluxDB PORT by env variable to tesla-history

mcbirse:
Added support to tesla-history script to define the InfluxDB PORT by environment variable for when running in a docker container per #183 

Jason - please push latest script build "0.1.1" to Docker Hub when you have chance, thanks!

Also fixed a few other issues I noticed, as follows:

* There was error output during execution of the setup script - the "type" command must output to stderr, so stderr has been redirected to null

```bash
./setup.sh: line 145: type: winpty: not found
```

* Changed the environment variable names for the tesla-history container to be more descriptive
* Added "user" (uid:gid) for tesla-history docker container as per how weather411 is configured
  * I had tried to edit the tesla-history.conf file, which was created during the setup process, however it was owned by "root" and could not be edited without first changing the owner
  * Perhaps the "user" configuration for tesla-history & weather411 should be configured via the compose.env file variable similar to grafana, instead of hard coding the uid:gid? e.g. ${GRAFANAUSER} - maybe even use the same variable?

### DIFF
--- a/tools/solar-only/setup.sh
+++ b/tools/solar-only/setup.sh
@@ -142,7 +142,7 @@ if [ -z "${TZ}" ]; then
     TZ="${CURRENT}"
 fi
 echo "Setup tesla-history..."
-if type winpty > /dev/null; then
+if type winpty > /dev/null 2>&1; then
     # Windows special case
     winpty docker exec -it -w /var/lib/tesla-history tesla-history python3 tesla-history.py --setup --timezone "${TZ}"
 else
--- a/tools/solar-only/powerwall.yml
+++ b/tools/solar-only/powerwall.yml
@@ -59,17 +59,19 @@ services:
             - influxdb
 
     tesla-history:
-        image: jasonacox/tesla-history:0.1.0
+        image: jasonacox/tesla-history:0.1.1
         container_name: tesla-history
         hostname: tesla-history
         restart: always
+        user: "1000:1000"
         volumes:
             - type: bind
               source: ./tesla-history
               target: /var/lib/tesla-history
         environment:
-            - IHOST=influxdb
-            - TCONF=/var/lib/tesla-history/tesla-history.conf
-            - TAUTH=/var/lib/tesla-history/tesla-history.auth
+            - INFLUX_HOST=influxdb
+            - INFLUX_PORT=8086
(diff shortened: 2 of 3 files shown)


tools/solar-only/setup.sh
```diff
@@ -142,7 +142,7 @@ if [ -z "${TZ}" ]; then
     TZ="${CURRENT}"
 fi
 echo "Setup tesla-history..."
-if type winpty > /dev/null; then
+if type winpty > /dev/null 2>&1; then
```
jasonacox, line 145: Great catch!

jasonacox:
Great updates @mcbirse !  The container `jasonacox/tesla-history:0.1.1` is pushed.

> Added "user" (uid:gid) for tesla-history docker container as per how weather411 is configured
I had tried to edit the tesla-history.conf file, which was created during the setup process, however it was owned by "root" and could not be edited without first changing the owner
Perhaps the "user" configuration for tesla-history & weather411 should be configured via the compose.env file variable similar to grafana, instead of hard coding the uid:gid? e.g. ${GRAFANAUSER} - maybe even use the same variable?

Good question. I struggle with the right answer here because we have so many different setups in the community (e.g. Synology, rootless, windows, *nix variants).  I want to accommodate as much as we can, but realize we should probably assert what we plan to support "out of the box" and differences will require users to customize themselves. In the context of your suggestion, I think the variable make sense. I suggest we push that change through our test suite to ensure it works without side effects. 

What would you say is our "assert to work" test suite?
* Standard Docker (not rootless - should visit to see how difficult it would be to pivot)
* Raspberry Pi 
* Ubuntu Linux 
* Windows 10/11 with WSL (complicated because the MS Linux support requires additional setup)
* MacOS
* No other InfluxDB, Grafana, Telegraf instances

And we could list, "can work with user customization":
* Rootless Docker
* Synology NAS
* Linux (outisde of Ubuntu LTS) or other *nix
* Shared InfluxDB 1.8/2.x, Grafana, Telegraf instances

This should probably move to a discussion... 😉

mcbirse:
Thanks Jason. It's complicated to support so many different setups however I feel the current stack is working quite well... it's those custom setups where it becomes difficult...

PR #289, opened 2023-06-05
add instruction: make backup script executable

s-crypt:
It took me a few minutes to figure out why I could not back up using the provided backup script (error `sudo: ./backup.sh: command not found`), but once I made it an executable it worked perfectly. 

I think adding an instruction for this would be helpful for some people (like me).


jasonacox:
Nice @s-crypt , that's a helpful addition!  Thanks! 🙏

s-crypt:
Thank you for making such an awesome and easy to use tool! Glad to be of any help!

Issue #286, opened 2023-06-01
Solar-only history is saving zero values when the API response has time gaps


youzer-name:
@mcbirse 

I've been running the solar-only history in a Docker container for about a week now. I noticed that yesterday and today there were zero kWh production values during times where there shouldn't have been.

Just now I saw this:

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/57046429/bb4e582f-3bc2-483b-a265-9fb075b2c8e4)

The yellow line is based on the tesla-history solar-only data. The orange line is the total solar production seen by my Gateway from both of my solar arrays. As you can see, the production didn't actually drop out. If it had, the orange line would also have dropped.

The raw influx data shows the zero point at 13:30Z:
```
2023-06-01T10:00:00Z 0         0       0    Jun   0     cloud  0       0     2023
2023-06-01T10:15:00Z 0         0       0    Jun   160   cloud  0       0     2023
2023-06-01T10:30:00Z 0         0       0    Jun   200   cloud  0       0     2023
2023-06-01T10:45:00Z 0         0       0    Jun   240   cloud  0       0     2023
2023-06-01T11:00:00Z 0         0       0    Jun   280   cloud  0       0     2023
2023-06-01T11:15:00Z 0         0       0    Jun   320   cloud  0       0     2023
2023-06-01T11:30:00Z 0         0       0    Jun   440   cloud  0       0     2023
2023-06-01T11:45:00Z 0         0       0    Jun   480   cloud  0       0     2023
2023-06-01T12:00:00Z 0         0       0    Jun   520   cloud  0       0     2023
2023-06-01T12:15:00Z 0         0       0    Jun   1240  cloud  0       0     2023
2023-06-01T12:30:00Z 0         0       0    Jun   2000  cloud  0       0     2023
2023-06-01T12:45:00Z 0         0       0    Jun   2040  cloud  0       0     2023
2023-06-01T13:00:00Z 0         0       0    Jun   2920  cloud  0       0     2023
2023-06-01T13:15:00Z 0         0       0    Jun   3280  cloud  0       0     2023
2023-06-01T13:30:00Z 0         0       0    Jun   0     cloud  0       0     2023
2023-06-01T13:45:00Z 0         0       0    Jun   3840  cloud  0       0     2023
2023-06-01T14:00:00Z 0         0       0    Jun   4040  cloud  0       0     2023
[...]

PR #293, opened 2023-06-10
Fix tesla daemon to replace data instead of update

mcbirse:
Update to tesla-history docker container to address issue noted by @youzer-name in #286 

@jasonacox - Please push to Docker Hub when you have a chance, thanks!

Changes as follows:

- Script will now remove previously written data before writing new data, instead of updating the data points in InfluxDB. This should address the issue where previously returned timestamps are not returned in subsequent calls for history data, which would result in zero values being seen in production graphs
- Updated the docker container to use Python 3.11 which is now released in alpine 3.18 (stable) - noting "Python 3.11 is between 10-60% faster than Python 3.10"

### DIFF
--- a/tools/solar-only/powerwall.yml
+++ b/tools/solar-only/powerwall.yml
@@ -59,7 +59,7 @@ services:
             - influxdb
 
     tesla-history:
-        image: jasonacox/tesla-history:0.1.1
+        image: jasonacox/tesla-history:0.1.2
         container_name: tesla-history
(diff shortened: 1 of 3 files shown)


jasonacox:
Merged and uploaded [jasonacox/tesla-history:0.1.2](https://hub.docker.com/repository/docker/jasonacox/tesla-history/general)

Thanks @mcbirse ! 🙏

PR #298, opened 2023-06-11
migrate more, fix y axis on monthy, tweaks

s-crypt:
* made all monthly graphs have a soft y = 0
* center y axis for Grid Usage
* migrated Wind, Pressure, and Precip to time series
* changed Solar Energy Year to all line graphs with fill for consistency (and to overlap), adjusted solar to 40% opacity to better see home usage
* removed invalid overrides for Self Powered pie chart


jasonacox:
I have to confess, the new style is growing on me... but still stuck on a few things:  1) I can't get excited about the new pie-graph, just not loving the skinny donut. 

old

image

new (skinny donut 😂 )

image



2) the Solar Energy Year home (blue) is too much clutter for the emphasis of the chart (solar) so I would turn the fill to Zero:

from 
image

to 
image

s-crypt:
I agree about the pie chart. What do you think about the pie view instead of the donut? 
image


I will create a new PR with the home usage line opacity + some other tweaks that I think make some improvements

jasonacox:
Hum... I don't think the full pie chart looks as "high tech" as the donut.  The original pie-chart uses the plugin instead of the built-in pie-chart.  Is there a reason you wanted to move away from the prior one? I understand it requires the plug-in but we are doing that for the others (boom table for example).

s-crypt:
I agree that it doesn't look as cool. The original pie chart plug in is marked as `(Old)` and grafana suggests migration to the new pie chart plugin. The reason I changed it was because I was migrating all of the old plugins to the new ones, pie chart included.

s-crypt:
I fixed the line opacity in #307

jasonacox:
Thanks @s-crypt - IMHO, the logic of replacing anything old is a good first step in an experiment, but ultimately we want to optimize for quality. The outcome (the look, features, function) of the dashboard is primary target. As I'm sure you have seen before, some updates are actually setbacks. I think the pie chart is one.  Sometimes you don't have a choice, other times you do.  In this case, I would prefer to keep the higher quality plugin, even if marked as old. I would also want to investigate forking the plugin code to see if we could contribute to the core to fix the built in grafana one... well, when we have time. :) 

On the time series updates, thanks for all you work on this! Originally, many of the features in the old time series panels were not ported to the new one. Over time I see that they have been added and as you have proven.  It seems we are now able to replicate all the features in the old panel with the new ones. There is still as aesthetic difference but it is growing on me. 😁

s-crypt:
Although I personally don't mind the new pie chart, you make some excellent points. Ill just use the new pie chart on my own dashboard and revert the template back to the old pie chart and merge that in to #307

PR #299, opened 2023-06-11
Partner files for v1.0 usage service

BuongiornoTexas:
Scripts, configuration files and example dashboards to use the pwdusage micro-service in the Powerwall-Dashboard stack.


jasonacox:
Thanks @BuongiornoTexas !

I was not able to get this to build for armv7 (RPi 3) due to numpy build failure during `pip install`, but it seems to work fine for arm64 (Apple Silicon + RPi 4) and amd64. 

I added my `upload.sh` script that I use to test and cross compile for architectures and upload to DockerHub.

jasonacox:
Uploaded container [jasonacox/pwdusage:1.0.0](https://hub.docker.com/repository/docker/jasonacox/pwdusage/general).

BuongiornoTexas:
> I added my upload.sh script that I use to test and cross compile for architectures and upload to DockerHub.

Thanks for this - getting into multi-arch builds is definitely out of my comfort zone. I'll update the README to make the docker build an optional step.

> I was not able to get this to build for armv7 (RPi 3) due to numpy build failure

I took a look around to see if I could find a solution to this and realised a) I don't think my docker install supports multiarch - at least not without additional work, so I can't test myself, and b) both numpy and pandas can have issues with installing into arm32/v7 docker images.
 
Unfortunately, I think sorting this one out is going to need someone who needs pwdusage for an RPi3 and is willing to spend the time resolving the problem. I'll raise it as an issue for pwdusage.

jasonacox:
Sounds good!  I agree on the armv7 approach.

PR #300, opened 2023-06-11
add and correct timezone references in queries

s-crypt:
Timezone in `Grid Status` for dashboard-new was hardcoded for new york (my fault for blindly copy and pasting)
Timezone is now added to the home and solar queries on `Solar Energy Year` on both the new and main dashboards, eliminating the offset that they had against the grid usage


jasonacox:
Great find, @s-crypt !!  I don't know how I missed this PR until now.  Thanks for the fix. 🙏

PR #301, opened 2023-06-12
Remove option to set email sender name in pwstatus

mcbirse:
Removed the option to set the e-mail sender name in the pwstatus script and config for better compatibility.

I noticed an issue when setting up a new test system, where the "-F" option of "mail" was no longer accepted to set the name of the e-mail sender (possibly due to a new version of postfix or mailx?).

I guess setting the FROM name of the sender probably should be left up to the user and their MUA/MTA/system setup instead.

The "mail" command used in the script is now generic and in the form of `mail -s "subject" [recipient-email-address]` only, which should be more generally accepted.


jasonacox:
Great fix!  Thanks @mcbirse !  🙏

PR #316, opened 2023-06-29
Dashboard-new formatting fixes and unlink library panels

s-crypt:
Also fix the savings panel on dashboard.json being too small compared to the other panels in the row


jasonacox:
Thanks @s-crypt !

PR #319, opened 2023-07-02
Fix dashboard.json in README

longzheng:
The `dashboard.json` was moved into the `/dashboards` folder.


jasonacox:
Good catch @longzheng !  Thank you! 🙏

PR #321, opened 2023-07-03
Update verify.sh to carry state of all tests back to calling shell.

vikrum:
I'm using `verify.sh` to check the health of the running instance periodically. A wrapping shell uses the `ALLGOOD` return value to trigger alerts.


jasonacox:
Thanks @vikrum !  Great addition. 🙏

PR #328, opened 2023-07-22
Proxy everything behind nginx

ssasso:
There are some situations (i.e., host behind Cloudflare Access/Tunnel) where you want to access all your services on a single http port.

This PR does:
* add a nginx container as rev proxy, and related config (see below for URL rewritings)
* update the Grafana dashboard for pointing to new info (using port from `window.location`) and URI
* update Grafana config to respond on `/grafana/`

## NGINX URL rules:

1. Proxy `/grafana/` to Grafana container
2. Send redirect for `/` to `/grafana/`
3. Map `/panel.html` to `/` of pypowerwall container
4. Proxy others `/*` to pypowerwall container


BJReplay:
Excellent work.  I couldn't get the powerwall panel working in my config behing nginx (I used NPM) - you've figured it out with your nginx config.  Nice one.

See https://github.com/jasonacox/Powerwall-Dashboard/tree/main/weather/contrib/ecowitt#readme for an example of using powerwall.extend.yml rather than editing powerwall.yml as a way of incorporating additional containers without modifying the base powerwall.yml - @jasonacox may prefer this approach.

jasonacox:
Great work @ssasso !  👍  And yes, I agree @BJReplay, it would be good to first roll this out as an optional add-on if we could. I would like to do some testing across all the different platforms (RPis, Ubuntu, MacOS, Windows, various NAS devices) before we make it the base install.

I'm not able to spend much time on the project the next few weeks, but would welcome community review and feedback on this PR.

Thanks everyone, and thanks for the PR @ssasso ! 🙏 

Jason

ssasso:
Thanks for the feedback @jasonacox and @BJReplay!
I can make the nginx container as an "extension", using `powerwall.extend.yml`.

However please note that also a Grafana config change is in place, to have it under `/grafana/`.
Luckily, Grafana people is smart, and automatically sends a redirect to `/grafana/` when the uri is `/` - so, some sort of backward compatibility is maintained even if you don't want to run the nginx rev proxy in front of the deployment.

@jasonacox if this can help, I am running this on RPi 4 (Raspbian 11), and Ubuntu 22.04 LTS. For sure Debian 11 will behave the same as raspbian, but I don't think we will have problems on other platforms, since all the changes are only on the containers (and I'm using the official nginx one)

Nordictrack:
I did the same thing and it worked!

Issue #330, opened 2023-07-24
Error in New PW Dashboard Solar Energy Year


aesculus:
I noticed that on the graph legend that the grid mean was really low compared to solar and home. It looks like its taking the mean over the entire year vs amounts that are > 0 like home and solar are doing.

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/283364/933d9723-eaf9-4aaf-bf78-784c44bafd58)

Also found this error when inspecting the query:

{
  "request": {
    "method": "GET",
    "url": "api/datasources/proxy/1/query",
    "params": {
      "db": "powerwall",
      "q": "SELECT sum(\"home\") FROM \"kwh\".\"http\" WHERE (\"home\" > 0) GROUP BY time(1d) fill(null) tz('America\\/Los_Angeles');SELECT sum(\"solar\") FROM \"kwh\".\"http\" WHERE (\"solar\" > 0) GROUP BY time(1d) fill(0) tz('America\\/Los_Angeles');SELECT sum(\"from_grid\") - sum(\"to_grid\") as \"grid\" FROM \"kwh\".\"http\" WHERE time >= now() - 24h and time <= now() GROUP BY time(1d) fill(0) tz('America\\/Los_Angeles')",
      "epoch": "ms"
    },
    "data": null,
    "precision": "ms",
    "hideFromInspector": false
  },
  "response": {
    "error": "error parsing query: found \\/, expected identifier, string, number, bool at line 1, char 98",
    "message": "error parsing query: found \\/, expected identifier, string, number, bool at line 1, char 98"
  }
}

jasonacox:
Hi @aesculus  - thanks for this.  Are you using this dashboard? 
https://github.com/jasonacox/Powerwall-Dashboard/blob/main/dashboards/dashboard.json

> I noticed that on the graph legend that the grid mean was really low compared to solar and home. It looks like its taking the mean over the entire year vs amounts that are > 0 like home and solar are doing.

Both solar and home should NOT have negative values so we remove any noise that may be caused by meters reporting negative flow.  However, for grid, you need to account for negative flow because you can push energy back to the grid. The "mean" will account for this so you could potentially have a very low or even negative mean.

> America\\/Los_Angeles

That's a typo.  The new dashboard pulls the timezone from the $tz grafana variable rather than have it hardcoded.  You might check to ensure it has the right variable value set.

[...]

PR #331, opened 2023-07-25
Fix mean in solar energy yr panel #330

jasonacox:


aesculus:
Don't forget to do the New Dashboard too unless it's built on top of this.

https://github.com/jasonacox/Powerwall-Dashboard/blob/main/dashboards/dashboard-new.json

jasonacox:
Ooof... good catch. I should remove that one.  I think we merged all the "new" aspects into the main.  I'll check before I do...

PR #333, opened 2023-07-29
Use templates for files depending on TZ value

rcasta74:
This is my proposal to use templates instead of to modify repository files to configure TZ.

All repository files that would be changed by `tz.sh` script have been renamed adding `.template`, and then `tz.sh` script has been changed to create them from templates using `envsubst` command. After this change, to upgrade to a new version, `git pull` command should be run before to run `upgrade.sh` script.

The `upgrade.sh` has been update to manage the case where it is started from an old version before this change: it will revert the files depending on TZ value before to make the pull (as usual), so that no conflicts will be raised, and then it recreates them from templates starting the new `tz.sh` script.

Files `VERSION` and `tz` have been removed from the repository and used only for local configuration.
The version is now retrieved with the command `git describe --tag --dirty=-custom` and saved in the file `VERSION` by `setup.sh` and `upgrade.sh` scripts to make a comparison with repository version during next updates.

In `upgrade.sh` script I have added a reference to a future tag `v2.9.11` supposing it will be added if this PR will be approved and merged: indeed it should be changed if the tag will be different.

Since I'm using `podman` instead of `docker`, I have not tested the changes to `setup.sh` script, while I have tested `tz.sh` script and the `upgrade.sh` script for the case the upgrade is started from an old version.
Sorry for this.


jasonacox:
Thanks @rcasta74  - I'll take a look.

jasonacox:
Hi @rcasta74  - I'm sorry for delay. 

I'm having some heartburn with the change. It makes sense, but need to still validate on all platforms with docker (RPi, Linux, MacOS, Win) and think through any unintended side effects that could be presented to any of the older version (for upgrade.sh). 

We will need to also resolve minor conflicts based to other changes, but those are minor.

If anyone else has validated this PR, please chime in here to help.

jasonacox:
Hi @rcasta74 — thanks again for putting this together. After sitting with it, I think the template approach adds complexity that makes onboarding harder for new users. I'm going to close this for now, but the underlying idea of making TZ handling cleaner is worth revisiting in a future dedicated effort. Thanks for the contribution!

PR #338, opened 2023-08-25
Shared crosshair

longzheng:
Screen Recording 2023-08-25 at 12 54 05 pm

I think it's useful to be able to see the correlation of different values on the different graphs at the same time.

Grafana has an [option](https://grafana.com/docs/grafana/latest/dashboards/build-dashboards/view-dashboard-json-model/) for "shared crosshair" which can show a crosshair on other timeline graphs when you hover over one

image

### DIFF
--- a/dashboards/dashboard.json
+++ b/dashboards/dashboard.json
@@ -125,7 +125,7 @@
   },
   "editable": true,
   "fiscalYearStartMonth": 0,
-  "graphTooltip": 0,
+  "graphTooltip": 1,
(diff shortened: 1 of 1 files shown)


BJReplay:
That is brilliant!

If you click on the gear icon near the top right, and then panel options, you can set the options for any custom dashboards you have as well.

image
image

jasonacox:
Wow! Love this @longzheng !  Thanks! 🙏

PR #354, opened 2023-09-20
Exit weather411 gracefully with SIGTERM

### LINKED ISSUE #353: python servers (pypowerwall and weather411) do not stop gracefully when container is stopped
I'm using `podman` so I don't know if it is the same with `docker`.

When I run `stop` command with `podman`, a `SIGTERM` signal is sent to the main process of the container, but it looks like that this signal is not handled properly within pypowerwall and weather411 containers since they do not stop and are killed after 10 seconds of timeout.
If I send them `SIGINT` signal instead, they stop gracefully.

rcasta74:
Fix #353

### DIFF
--- a/weather/server.py
+++ b/weather/server.py
@@ -77,6 +77,7 @@ import json
 import requests
 import resource
 from datetime import datetime
+import signal
 import sys
 import os
 from http.server import BaseHTTPRequestHandler, HTTPServer, ThreadingHTTPServer
@@ -426,6 +427,9 @@ def api(port):
             print(' CANCEL \n')
     sys.stderr.write('\r ! apiServer Exit\n')
 
+def sigTermHandler(signum, frame):
+    sys.exit()
+
 # MAIN Thread
 if __name__ == "__main__":
     # Create threads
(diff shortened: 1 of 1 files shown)


weather/server.py
```diff
@@ -426,6 +427,9 @@ def api(port):
             print(' CANCEL \n')
     sys.stderr.write('\r ! apiServer Exit\n')
 
+def sigTermHandler(signum, frame):
+    sys.exit()
```
mcbirse: My recommendation/preference would be to use `raise SystemExit` per explanation [here](https://github.com/jasonacox/Powerwall-Dashboard/issues/353?notification_referrer_id=NT_kwDOBnsFf7Q3NzcxNzQzMTA2OjEwODcyNTYzMQ#issuecomment-1728709935)

mcbirse (approved):

mcbirse (approved):

jasonacox:
Thank you @rcasta74 ! This is a great enhancement.

Thanks @mcbirse !

jasonacox:
I have pushed weather411 and pypowerwall updates. Testing looks good. The `upgrade.sh` will pull the new versions.

Thanks again! 🙏

PR #366, opened 2023-10-10
Update tesla-history with retry handling on errors

mcbirse:
Hi Jason,

This PR includes various updates for the Tesla-History script and Docker Compose environment. Currently for the [Solar-Only Dashboard](https://github.com/jasonacox/Powerwall-Dashboard/issues/183) #183 but with plans to migrate into the main project all going well.

I have not committed my own changes or merged this PR as the new tesla-history script will need to be pushed to the Docker Hub account please when you can.

Updates and improvements to tesla-history script:

- add retry handling on errors, which should resolve issues noted by @youzer-name in #356 
- add signal handler to exit gracefully on SIGTERM, similar to #353 noted by @rcasta74 
- always set config/variables from environment if defined (instead of just when in daemon mode)
- the "--remove" option no longer requires login to Tesla account to be used
- update setup script to always restart tesla-history container, since existing config could be modified when re-running setup
- change to `Dockerfile` to fix issue with missing build dependencies (i.e. gcc) when building & pushing container to Docker Hub

Docker Compose config changes:

- Update `powerwall.yml` to use variables for "user" and "ports" in containers, per #357 and #360 noted by @hulkster
- Updated `compose.env.sample` with explanation of latest supported options
- Proposing change to `powerwall.yml` to use "unless-stopped" as the default restart policy for containers going forward (testing/research shows this may be a better choice, happy to discuss/explain my reasoning)

@jasonacox - Please push the latest tesla-history script to Docker Hub by running `upload.sh`

Please note, I struck some issues while testing this in my dev environment.

First issue - my docker buildx environment no longer worked for some reason (something broken with latest buildx failing with cgroup errors). 

I found a workaround which was to install a different buildx driver, as below:

```bash
# Install fixed buildx instance using rootless driver image
docker buildx create --name fixed_builder --driver-opt 'image=moby/buildkit:v0.12.1-rootless' --bootstrap --use
```

I'm not sure if you will need to do the same for your environment.

Second issue - note the "RUN" command in `Dockerfile` has been been updated.

The build was failing due a dependency requiring gcc to be built. I found a nice solution which was to install alpine linux's "build-base" package (includes gcc) before running "pip install", and then it deletes the package afterwards, which keeps the container size small. Works a charm.

Please note this PR is the first with some more to come. I am working on merging the entire solar-only offshoot into the main project.

I have some preliminary changes in the works, with a more modular approach to how we handle the required docker containers, which will be determined during the setup/install process based on the user choosing a configuration option (default or solar-only).

The containers which will be installed/run are then defined by using the profiles feature of Docker Compose. Testing so far looks promising, but this is still a work in progress at the moment.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -20,7 +20,7 @@ echo "-----------------------------------------"
 if [ "$EUID" -eq 0 ]; then 
   echo "ERROR: Running this as root will cause permission issues."
   echo ""
-  echo "Please ensure your local user in in the docker group and run without sudo."
+  echo "Please ensure your local user is in the docker group and run without sudo."
   echo "   sudo usermod -aG docker \$USER"
   echo "   $0"
   echo ""
--- a/tools/solar-only/compose.env.sample
+++ b/tools/solar-only/compose.env.sample
@@ -19,6 +19,9 @@ GRAFANAUSER="1000:1000"
 # Accepts a colon-separated string to set the host IP, host port, and
 # container port in the form [[IP:]HOST_PORT:]CONTAINER_PORT[/PROTOCOL]
 #
+# Use 127.0.0.1 (or ::1 for IPV6) to prevent services from being exposed
+# outside of your local machine - e.g. INFLUXDB_PORTS="127.0.0.1:8086:8086"
+#
 #INFLUXDB_PORTS="8086:8086"
 #GRAFANA_PORTS="9000:9000"
 #WEATHER411_PORTS="8676:8676"
--- a/tools/solar-only/compose.env.sample
+++ b/tools/solar-only/compose.env.sample
@@ -4,3 +4,21 @@
 # See issue #22,
 # https://github.com/jasonacox/Powerwall-Dashboard/issues/22#issuecomment-1253076441
 GRAFANAUSER="1000:1000"
+
+#------------------------------------------------------------------------------#
+# Powerwall-Dashboard user and group (uid:gid)
+#------------------------------------------------------------------------------#
+# User and group dashboard was installed under and containers will run as
+# (uncomment and modify if permission issues occur due to non-default uid/gid).
+#
+#PWD_USER="1000:1000"
+
+#------------------------------------------------------------------------------#
+# Container ports to use (uncomment and modify only if custom ports required)
+#------------------------------------------------------------------------------#
+# Accepts a colon-separated string to set the host IP, host port, and
+# container port in the form [[IP:]HOST_PORT:]CONTAINER_PORT[/PROTOCOL]
+#
+#INFLUXDB_PORTS="8086:8086"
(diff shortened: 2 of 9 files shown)


hulkster:
Thanks for all your work @mcbirse 

I'm wondering if maybe worth adding a comment in compose.env.sample along the lines of:
    _"Setting IP to 127.0.0.1 (or ::1 for IPV6) will prevent these services from being exposed outside of your local machine"_
This makes it a little more clear what you can use that for and why.

setup.sh
```diff
@@ -20,7 +20,7 @@ echo "-----------------------------------------"
 if [ "$EUID" -eq 0 ]; then 
   echo "ERROR: Running this as root will cause permission issues."
   echo ""
-  echo "Please ensure your local user in in the docker group and run without sudo."
+  echo "Please ensure your local user is in the docker group and run without sudo."
```
jasonacox, line 23: Wow, great catch. How long has that been there. 😊

jasonacox:
> I'm wondering if maybe worth adding a comment in compose.env.sample along the lines of: _"Setting IP to 127.0.0.1 (or ::1 for IPV6) will prevent these services from being exposed outside of your local machine"_ This makes it a little more clear what you can use that for and why.

Perhaps list an example? e.g.
```
# Use 127.0.0.1 (or ::1 for IPV6) to prevent services from being exposed outside of 
# your local machine - ex. INFLUXDB_PORTS="127.0.0.1:8086:8086"
```

mcbirse:
> > I'm wondering if maybe worth adding a comment in compose.env.sample along the lines of: _"Setting IP to 127.0.0.1 (or ::1 for IPV6) will prevent these services from being exposed outside of your local machine"_ This makes it a little more clear what you can use that for and why.
> 
> Perhaps list an example? e.g.
> 
> ```
> # Use 127.0.0.1 (or ::1 for IPV6) to prevent services from being exposed outside of 
> # your local machine - ex. INFLUXDB_PORTS="127.0.0.1:8086:8086"
> ```

Great suggestion @hulkster and @jasonacox - I have updated the `compose.env.sample` file and added the example.

setup.sh
mcbirse: Lol, yes never noticed this until recently... didn't pick up on it before when I was looking for spelling errors, since the spelling is good! 🤣

jasonacox:
@mcbirse - Build and Push jasonacox/tesla-history to DockerHub

Version: 0.1.3
* VERIFY jasonacox/tesla-history:latest
  Platform:  linux/amd64
  Platform:  linux/arm64
  Platform:  linux/arm/v7

* VERIFY jasonacox/tesla-history:0.1.3
  Platform:  linux/amd64
  Platform:  linux/arm64
  Platform:  linux/arm/v7

mcbirse:
Thanks Jason!

Issue #373, opened 2023-10-15
Update Versions of Grafana and Telegraf


jasonacox:
This is a proposal to upgrade the pinning to newer version of Grafana and Telegraf:

* Grafana (v9.4.14)
* Telegraf (v1.28.2)
* pyPowerwall (v0.6.2t28) - Required to support background color adjustments for newer Grafana versions

During testing, the "look" difference between Grafana v9.1.2 and v9.4.14 are minor. There is a subtle adjustment to the font and labeling (IMHO looks better). No issues with panels. The Power Flow Animation required an update to pyPowerwall to accommodate the grafana-dark background color (transparent settings were in use before, but no longer work).

Related discussions:
* https://github.com/jasonacox/Powerwall-Dashboard/discussions/371
* https://github.com/jasonacox/Powerwall-Dashboard/discussions/363
* https://github.com/jasonacox/Powerwall-Dashboard/issues/368

rcasta74:
Why not use Alpine-based images?
```
docker.io/grafana/grafana        9.4.14                f69ac1c180a5  7 weeks ago    332 MB
docker.io/grafana/grafana        9.4.14-ubuntu         76a0cce050ec  2 months ago   428 MB
```
```
docker.io/library/telegraf       1.28.2-alpine     3d231738b7db  2 weeks ago    210 MB
docker.io/library/telegraf       1.28.2            f10f8ba7295f  3 weeks ago    453 MB
```

jasonacox:
Good question. Is there a reason? Key is to make sure we target all architectures (amd64, arm32, arm64) to ensure it they have builds that run on things from Raspberry Pis, x86 to Macs.

BuongiornoTexas:
> Good question. Is there a reason? 

I know I ran into an issue when I was doing my ToU service - alpine does not include gcc (required by pandas). Not sure if this could affect any of the core modules in the dashboard. I also have a memory that there is an issue with RPi support for some of the python versions.

jasonacox:
Latest release moves us to Grafana 12. Closing this for now.

Issue #368, opened 2023-10-11
Docker Desktop Upgrade issue with Telegraf


jaffro2k:
Hi folks - I've had an issue where Dock Desktop upgraded to v4.24.1 (Windows 11) - and seems to have broken Telegraf.

The container is now stuck in a restart loop with the log showing this error:  

Loading config file: /etc/telegraf/telegraf.conf
error loading config file /etc/telegraf/telegraf.conf: open /etc/telegraf/telegraf.conf: operation not permitted

I subsequently tried to run upgrade.sh to see if there was an update to Telegraf however same issue.

Any ideas?  Would like to fix rather than re-building again from scratch (and ideally keep the history in the DB).

Cheers,

Jeff

jasonacox:
Hi @jaffro2k !

First, before you do anything else, backup your data:

```bash
# stop the containers
cd Powerwall-Dashboard # in case you aren't there already
./compose-dash.sh stop

# create a backup
tar -zcvf influxdb-snapshot.tgz influxdb 
```

I recommend you copy that `influxdb-snapshot.tgz` file to a safe place.

Second, I'm not a Windows 11 user so I would welcome any help on the "cause".  Having said that, if this is a known issue, telegraf may have fixed it in an update. The `upgrade.sh` won't fix it because we pin to a version.  Here is a way to test:

```bash
# first back up the config
cp powerwall.yml powerwall.yml.bak
```
Edit the powerwall.yml file and update the version next to telegraf from 1.26.1 to 1.28.2:

```yml
    telegraf:
        image: telegraf:1.26.1
        container_name: telegraf
```

to

```yml
    telegraf:
        image: telegraf:1.28.2
        container_name: telegraf
```

Now try to start the stack:

```bash
./compose-dash.sh up -d
```

Check telegraf to see what happens: `docker logs telegraf`

jaffro2k:
Thanks @***@***.***>

I went through the steps and looks like an updated telegraf version was pulled down – however I still have the same error with the container stuck in a restarting mode:

2023-10-12 19:16:14 2023-10-12T08:16:14Z I! Loading config: /etc/telegraf/telegraf.conf
2023-10-12 19:16:14 2023-10-12T08:16:14Z E! error loading config file /etc/telegraf/telegraf.conf: open /etc/telegraf/telegraf.conf: operation not permitted

Any other thoughts?

Regards,

Jeff

jasonacox:
[...]

PR #374, opened 2023-10-15
V2.10.0 - Docker Compose and Container Upgrades

jasonacox:
# v2.10.0 - Updates

## Docker Compose Config Improvements 
- Update `powerwall.yml` to use variables for "user" and "ports" in containers, per #357 and #360 noted by @hulkster
- Updated `compose.env.sample` with explanation of latest supported options
- Updated `powerwall.yml` to use "unless-stopped" as the default restart policy for containers going forward
- Credit: @mcbirse - ref #366 

## Upgrade Containers
* This PR will update the pinning to newer versions of Telegraf and pyPowerwall in `powerwall.yml`: Telegraf (v1.28.2), pyPowerwall (v0.6.2t28)
* `setup.sh` - Adds PW_STYLE to pypowerwall.env settings to support newer Grafana versions.
* `upgrade.sh` - Some cleanup and handling for new PW_STYLE for existing installations.

## Related:
* #366
* #373 
* https://github.com/jasonacox/Powerwall-Dashboard/discussions/371
* https://github.com/jasonacox/Powerwall-Dashboard/discussions/363
* https://github.com/jasonacox/Powerwall-Dashboard/issues/368


mcbirse:
Hi Jason,

I have added commits with the proposed changes for `powerwall.yml` (similar to #366) which include:

- Update `powerwall.yml` to use variables for "user" and "ports" in containers
- Update `compose.env.sample` with explanation of latest supported options
- Update `upgrade.sh` script to replace GRAFANAUSER with PWD_USER in existing `compose.env` files, since this is now used for all containers including Grafana
- Change to `powerwall.yml` to use "unless-stopped" as the default restart policy for containers going forward (testing/research shows this may be a better choice, happy to discuss/explain my reasoning)

Also:

- Update other Docker Hub upload scripts to include your verification of specific version
- Add optional 3rd argument to `compose-dash.sh` script, which is useful sometimes to pass additional arguments to docker compose (e.g. being able run "compose-dash.sh up -d --remove-orphans" due to docker compose container changes could be handy)

Please review, and happy for you to modify as you see fit, or let me know if you see any issues with the changes.

upgrade.sh
```diff
@@ -113,13 +113,33 @@ fi
 # Silently create default docker compose env file if needed.
 if [ ! -f ${COMPOSE_ENV_FILE} ]; then
     cp "${COMPOSE_ENV_FILE}.sample" "${COMPOSE_ENV_FILE}"
+else
+    # Convert GRAFANAUSER to PWD_USER in existing compose env file (required in 2.10.0)
+    sed -i "s@GRAFANAUSER@PWD_USER@g" "${COMPOSE_ENV_FILE}"
+    if grep -q "^PWD_USER=\"1000:1000\"" "${COMPOSE_ENV_FILE}"; then
+        sed -i "s@^PWD_USER=\"1000:1000\"@#PWD_USER=\"1000:1000\"@g" "${COMPOSE_ENV_FILE}"
+    fi
```
jasonacox, line 121: Nice!  Here is what this does:

* It replaces all occurrences of "GRAFANAUSER" with "PWD_USER" in the `compose.env` file.
* It checks if a line starting with "PWD_USER=" followed by "1000:1000" exists in the file. If it does, it comments out that line by adding a "#" at the beginning. This is because the `powerwall.env` will default to that.

compose-dash.sh
```diff
@@ -52,14 +52,14 @@ else
     pwextend=""
 fi
 
-echo "Running Docker Compose..."  
+echo "Running Docker Compose..."
 if docker-compose version > /dev/null 2>&1; then
     # Build Docker (v1)
-    docker-compose -f powerwall.yml $pwextend $1 $2
+    docker-compose -f powerwall.yml $pwextend $1 $2 $3
```
jasonacox, line 58: Thanks @mcbirse  - This makes sense. In fact, had me wonder if we should just include all arguments. Should we just use `$@` which would be all arguments?

jasonacox:
GREAT additions @mcbirse!  Thank you! 🙏  I'm going to start testing this before we merge...

jasonacox:
I'm seeing some issues with the Power Flow animation where a light gray border appears around the animation. I need to investigate...

image

I'm seeing JavaScript errors as the resize for the animation happens. 

image

I'm going to try different Grafana versions to see if this is a bug or a conflict with our code. If it is a conflict, I'll likely revert the Grafana upgrade with the `PW_STYLE=clear` option in pypowerwall.env for anyone who wants to experiment with newer Grafana versions to get it to work.

compose-dash.sh
mcbirse: Hi Jason, I agree, it would probably make sense to simply use `$@` so there is no limit on the number of additional arguments that could be passed. Please feel free to update it.

upgrade.sh
mcbirse: Yep - all defaults are set in the `powerwall.yml` docker compose file now when a variable is undefined.

jasonacox:
I was finally able to get a clean run (no JavaScript errors) with Grafana v9.2.20. The border artifacts go away and the animation looks good with `PW_STYLE=clear` and `PW_STYLE=grafana-dark`.   However, pushing it to recompute based on date ranges and other standard tasks I would do, did produce some JavaScript errors but not related to the animation. 

From my testing (one version at a time) I discovered that something changed with v9.3+ that causes problems with our iFrame animation code and auto-resizing, specifically the border artifacts and JS errors. At v9.5 a different panel title formatting is introduced that seems to break the look as well. I think it will require a lot more investigation.

I'm reverting the upgrade of Grafana and keeping it at v9.1.2 as it seems the most stable in my testing. It also works well with both `clear` and `grafana-dark` modes. I'm going to continue to investigate how we get to a newer version. The other updates make sense to push forward and having `grafana-dark` as the default will help any others exploring newer Grafana version, including the ones raised in the Issues and Discussions.

jasonacox:
v2.10.0 has been released....

* Tested Upgrade - no issues
* Tested Clean Install - no issues

image

jgleigh:
@jasonacox Getting this error while trying to upgrade:
sed: 1: "compose.env": command c expects \ followed by text

mcbirse:
Hey @jgleigh - I have pushed a quick fix to the upgrade script which should hopefully fix this problem. 🤞 

I looked into this error, and it appears to be a compatibly issue between GNU sed (Linux based) and BSD sed (used in Mac OS). Are you using Mac?

It seems `sed -i` (update in-place with no backup file) does not work on Mac. Instead it expects a backup file, whereas on Linux this is accepted and will skip creating the backup file.

`sed -i''` would be acceptable on both platforms to skip creating the backup file. However I decided to change the command to `sed -i.bak` to create the backup file anyway, as this is consistent with use of sed throughout the other scripts so must work across all platforms (hopefully).

Let us know if the upgrade works for you now.

jgleigh:
@mcbirse That fixed it! Yes this was on a Mac. Thanks again.

jasonacox:
Thanks @jgleigh for spotting the issue and thank @mcbirse for the fix!

PR #386, opened 2023-11-18
V3.0.0 - Integrate Solar Only Support

mcbirse:
# v3.0.0 - Updates

## Integrate Solar Only Support

* Added Solar Only support as a setup option when installing Powerwall Dashboard by @mcbirse
* Updated Solar Only dashboard with user adjustable tz and cost variables, and removed Grid Status & Current State panels
* Updated Docker Compose environment to support compose profiles and updated setup, upgrade and verify scripts
* Added checks to setup script for common issues such as user/group problems
* Added migration for beta Solar Only installs to upgrade script

## Related
* https://github.com/jasonacox/Powerwall-Dashboard/issues/183
* https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools/solar-only


dashboards/dashboard.json
```diff
@@ -1439,7 +1439,7 @@
             ]
           ],
           "tags": [],
-          "tz": "America/Los_Angeles"
+          "tz": "$tz"
```
jasonacox, line 1442: Wow, good catch. This is a bug fix. ;)

setup.sh
```diff
@@ -50,6 +120,80 @@ if ! docker-compose version > /dev/null 2>&1; then
...
+else
+    # Prompt for configuration profile if not set
+    echo "Select configuration profile:"
+    echo ""
+    echo " 1 - default     (Powerwall w/ Gateway on LAN)"
+    echo " 2 - solar-only  (No Gateway - data retrieved from Tesla Cloud)"
+    echo ""
+    while :
+    do
+        read -r -p "Select profile: " response
```
jasonacox, line 137: Minior thought: I wonder if we make the default 1 by also "Select profile [1]: " which means someone leaving it blank (enter) will get the default. We do that for other defaults.

setup.sh
```diff
@@ -50,6 +120,80 @@ if ! docker-compose version > /dev/null 2>&1; then
...
+    if [ "${PWD_USER}" == "1000:1000" ]; then
+        desc="normal default"
+        name="Default"
+    else
+        desc="configured user"
+        name="  Saved"
+    fi
+    CURRENT_USER="$(id -u):$(id -g)"
+    if [ "${CURRENT_USER}" != "${PWD_USER}" ]; then
+        echo "WARNING: Your current user uid/gid does not match the ${desc}."
```
jasonacox, line 168: Love this!

upgrade.sh
```diff
@@ -24,6 +26,12 @@ if [ "$EUID" -eq 0 ]; then
     exit 1
 fi
 
+# Windows Git Bash docker exec compatibility fix
+if type winpty > /dev/null 2>&1; then
+    shopt -s expand_aliases
+    alias docker="winpty -Xallow-non-tty -Xplain docker"
```
jasonacox, line 32: I'm glad you discovered/added this!

verify.sh
```diff
@@ -125,39 +200,46 @@ CONTAINER="telegraf"
...
-echo -e "${dim} - Version: ${subbold}$VER"
 echo -e ""
 
 # influxdb
 echo -e "${bold}Checking influxdb${dim}"
 echo -e "----------------------------------------------------------------------------"
 CONTAINER="influxdb"
 VER=$UKN
-CONF_FILE="pypowerwall.env"
+CONF_FILE="influxdb.conf"
```
jasonacox, line 242: Another good find/bug fix!

jasonacox:
✅  Checking Upgrade on RPi Setup from 2.10.0 to 3.0.0 - Steps to force upgrade:

```bash
git pull
gh pr checkout 386
cp upgrade.sh tmp.sh
bash tmp.sh upgrade
...
./verify.sh
```
Output
```
WARNING: You already have the latest version (v3.0.0).

Upgrade Powerwall-Dashboard from 3.0.0 to 3.0.0
---------------------------------------------------------------------
This script will attempt to upgrade you to the latest version without
removing existing data. A backup is still recommended.

Upgrade - Proceed? [y/N] Y

Resetting Timezone to Default...

Pull influxdb.sql, dashboard.json, telegraf.conf, and other changes...

No local changes to save
Already up to date.


Setting Timezone back to America/Los_Angeles...

Updating Powerwall-Dashboard stack...
Running Docker Compose...
[+] Running 5/5
 ✔ Container influxdb     Running                                                                                                                  0.0s 
 ✔ Container pypowerwall  Started                                                                                                                  3.5s 
 ✔ Container weather411   Started                                                                                                                  3.8s 
 ✔ Container grafana      Running                                                                                                                  0.0s 
 ✔ Container telegraf     Started                                                                                                                  3.2s 

Waiting for InfluxDB to start...
 up!

Add downsample continuous queries to InfluxDB...
2023/11/18 17:47:35 error: retention policy already exists
2023/11/18 17:47:35 error: retention policy already exists
2023/11/18 17:47:35 error: retention policy already exists
2023/11/18 17:47:35 error: continuous query already exists
2023/11/18 17:47:35 error: continuous query already exists

Deleting old pyPowerwall...
pypowerwall
pypowerwall

Deleting old telegraf...
telegraf
telegraf

Deleting old weather411...
weather411
weather411

Restarting Powerwall-Dashboard stack...
Running Docker Compose...
[+] Running 5/5
 ✔ Container pypowerwall  Started                                                                                                                  2.0s 
 ✔ Container influxdb     Running                                                                                                                  0.0s 
 ✔ Container grafana      Running                                                                                                                  0.0s 
 ✔ Container weather411   Started                                                                                                                  2.3s 
 ✔ Container telegraf     Started                                                                                                                  2.0s 

---------------[ Update Dashboard ]---------------
Open Grafana at http://localhost:9000/

From 'Dashboard/Browse', select 'New/Import', and
upload 'dashboard.json' located in the folder
/home/pi/Powerwall-Dashboard/dashboards/

Please note, you may need to select data sources
for 'InfluxDB' and 'Sun and Moon' via the
dropdowns and use 'Import (Overwrite)' button.
```
```
Verify Powerwall-Dashboard 3.0.0 on Linux - Timezone: America/Los_Angeles


Checking configuration
 - Dashboard configuration: default
 - EnvVar COMPOSE_PROFILES: default,weather411

Checking pypowerwall
 - Service (port 8675): GOOD
 - Version: 0.6.2 Proxy t28
 - Powerwall State: CONNECTED - Firmware Version: 23.28.2

Checking telegraf
 - Container (telegraf): GOOD

Checking influxdb
 - Container (influxdb): GOOD
 - Service (port 8086): GOOD

Checking grafana
 - Container (grafana): GOOD
 - Service (port 9000): GOOD

Checking tesla-history
 - Skipped: Only required in 'solar-only' configuration

Checking weather411
 - Config File weather/weather411.conf: GOOD
 - Service (port 8676): GOOD
 - Weather: {"temperature": 15.86}
 - Version: 0.2.2

All tests succeeded.
```

setup.sh
jasonacox: Also, the way the logic runs here, it will only prompt for configuration change (default or solar-only) the first time you run setup. I wonder if we should have a condition prompt like, "You are using the Powerwall (vs Solar-only) configuration, do you want to change that? [y/N]" to allow someone to switch. I suspect the switch would likely only be caused by an error running it first and selecting the wrong one, or if someone with Solar adds Powerwalls. What do you think?

jasonacox: After a few more test and tweaking my idea, I'm leaning towards just leaving it as you have it. Keeping it simple for v3.0.0 makes the most sense. I'm going to test a bit more and then accept the PR if all looks well.

jasonacox:
Issue with a Ubuntu hosted instance during upgrade:

```
Upgrade Powerwall-Dashboard from 2.10.0 to 3.0.0
---------------------------------------------------------------------
This script will attempt to upgrade you to the latest version without
removing existing data. A backup is still recommended.

Upgrade - Proceed? [y/N] y

Resetting Timezone to Default...

Pull influxdb.sql, dashboard.json, telegraf.conf, and other changes...

No local changes to save
remote: Enumerating objects: 111, done.
remote: Counting objects: 100% (111/111), done.
remote: Compressing objects: 100% (54/54), done.
remote: Total 111 (delta 68), reused 94 (delta 54), pack-reused 0
Receiving objects: 100% (111/111), 60.95 KiB | 770.00 KiB/s, done.
Resolving deltas: 100% (68/68), completed with 14 local objects.
From https://github.com/jasonacox/Powerwall-Dashboard
   6e26eee..faa3a81  main       -> origin/main
 * [new branch]      v3.0.0     -> origin/v3.0.0
Updating 6e26eee..faa3a81
Fast-forward
 .gitignore                                                 |    7 +
 README.md                                                  |   61 +-
 RELEASE.md                                                 |    8 +
 VERSION                                                    |    2 +-
 WINDOWS.md                                                 |  201 +++
 compose.env.sample                                         |   11 +-
 dashboards/README.md                                       |   16 +-
 .../dashboard.json => dashboards/dashboard-solar-only.json |  574 +++------
 dashboards/dashboard.json                                  |    4 +-
 powerwall.yml                                              |   26 +
 setup.sh                                                   |  286 ++++-
 tools/README.md                                            |   36 +-
 tools/solar-only/README.md                                 |  130 +-
 tools/solar-only/compose-dash.sh                           |   69 --
 tools/solar-only/compose.env.sample                        |   27 -
 tools/solar-only/grafana.env.sample                        |   34 -
 tools/solar-only/grafana/.keep                             |    1 -
 tools/solar-only/influxdb.conf                             |  154 ---
 tools/solar-only/influxdb/dropcq.sql                       |   25 -
 tools/solar-only/influxdb/influxdb.sql                     |   42 -
 tools/solar-only/influxdb/run-once-2.8.0.sql               |   15 -
 tools/solar-only/influxdb/run-once-2.8.3.sql               |   13 -
 tools/solar-only/influxdb/run-once-2.9.2.sql               |   18 -
 tools/solar-only/powerwall.yml                             |   71 --
 tools/solar-only/setup.sh                                  |  178 ---
 tools/solar-only/tesla-history/README.md                   |  286 -----
 tools/solar-only/tesla-history/tesla-history.py            | 1560 ------------------------
 tools/solar-only/tz                                        |    1 -
 tools/solar-only/tz.sh                                     |   31 -
 tools/solar-only/weather.sh                                |  149 ---
 tools/solar-only/weather/weather411.conf.sample            |   34 -
 tools/{solar-only => }/tesla-history/Dockerfile            |    0
 tools/tesla-history/tesla-history.py                       |  793 +++++++++---
 tools/{solar-only => }/tesla-history/upload.sh             |    0
 tz.sh                                                      |   18 +-
 upgrade.sh                                                 |  247 +++-
 verify.sh                                                  |  355 ++++--
 weather.sh                                                 |   86 +-
 38 files changed, 1964 insertions(+), 3605 deletions(-)
 create mode 100644 WINDOWS.md
 rename tools/solar-only/dashboard.json => dashboards/dashboard-solar-only.json (85%)
 delete mode 100755 tools/solar-only/compose-dash.sh
 delete mode 100644 tools/solar-only/compose.env.sample
 delete mode 100644 tools/solar-only/grafana.env.sample
 delete mode 100644 tools/solar-only/grafana/.keep
 delete mode 100644 tools/solar-only/influxdb.conf
 delete mode 100644 tools/solar-only/influxdb/dropcq.sql
 delete mode 100644 tools/solar-only/influxdb/influxdb.sql
 delete mode 100644 tools/solar-only/influxdb/run-once-2.8.0.sql
 delete mode 100644 tools/solar-only/influxdb/run-once-2.8.3.sql
 delete mode 100644 tools/solar-only/influxdb/run-once-2.9.2.sql
 delete mode 100644 tools/solar-only/powerwall.yml
 delete mode 100755 tools/solar-only/setup.sh
 delete mode 100644 tools/solar-only/tesla-history/README.md
 delete mode 100644 tools/solar-only/tesla-history/tesla-history.py
 delete mode 100644 tools/solar-only/tz
 delete mode 100755 tools/solar-only/tz.sh
 delete mode 100755 tools/solar-only/weather.sh
 delete mode 100644 tools/solar-only/weather/weather411.conf.sample
 rename tools/{solar-only => }/tesla-history/Dockerfile (100%)
 rename tools/{solar-only => }/tesla-history/upload.sh (100%)
 mode change 100644 => 100755
Current branch main is up to date.


Setting Timezone back to America/Los_Angeles...

Updating Powerwall-Dashboard stack...
Running Docker Compose...
ERROR: The Compose file './powerwall.yml' is invalid because:
Unsupported config option for services.pypowerwall: 'profiles'
Unsupported config option for services.telegraf: 'profiles'
Unsupported config option for services.tesla-history: 'profiles'
Unsupported config option for services.weather411: 'profiles'
```

jasonacox:
Issue corrected with v3.0.1.

jasonacox:
Thanks for this upgrade @mcbirse !!! 🙏 

Please feel free to propose an alternative to my v3.0.1 patch. I need to add the docker-compose V1 hosts to my testing plan going forward. In any case, it seems to work. Looking for any input from others.

mcbirse:
Hi @jasonacox - I committed directly a bug fix for the v3.0.1 patch as I believe the order in which the checks for docker compose V1/V2 need to be the other way round.

i.e. check for V2 first using `docker compose version` and if that fails, then fall back to the V1 check `docker-compose version` and use the individual .yml files since "profiles" is likely not supported.

Why check for V2 first? 

The `docker-compose` command is aliased when Docker Compose V2 is installed (at least it is on my system, and it is also mentioned in the docs: https://docs.docker.com/compose/migrate/#how-do-i-switch-to-compose-v2)

Running both commands on my Alpine Linux system produces the below:

```bash
$ docker-compose version
Docker Compose version v2.17.3
```

```bash
$ docker compose version
Docker Compose version v2.17.3
```

So running the first command (with the dash) is not a definitive test to say "only Docker Compose V1 is installed", so we should check for V2 first.

This is just a quick fix, hopefully there are no issues with it.

I would still like to look further for other alternatives to the patch to support Docker Compose V1 users, however the current solution is good I believe.

setup.sh
mcbirse: I think this may help to solve a lot of potential issues, hopefully.

I tested on some different systems and ensured I was installing with a different uid/gid, and having this a prompt during setup to configure the system for that user solved those permission issues (or at the least will highlight the potential issues to the user).

upgrade.sh
mcbirse: Results of my adventure down the Windows rabbit hole....

setup.sh
mcbirse: As it happens, I had originally written the setup as you described, with a default on the "Select profile" prompt if the user pressed enter, and also allowing the chosen profile to be changed by re-running setup...

I later changed my thinking about this, and went with a forced prompt (user MUST make a selection, to prevent accidentally pressing enter and choosing the config profile without consciously looking at the options and making a selection). And also, NOT allowing the chosen profile to be changed later, as swapping and changing did not seem sensible (and could be achieved if you really wanted to by manually editing `compose.env`).

If we wanted to pivot to this method in future I had written some prelim methods to handle it as below:

```bash
# Read list of docker compose profiles from env file
. "${COMPOSE_ENV_FILE}"
IFS=',' read -a PROFILES <<< ${COMPOSE_PROFILES}
PROFILE="none"
current="1"
changed=0
index=-1
for i in "${!PROFILES[@]}"; do
    if [[ "${PROFILES[$i]}" == "default" ]]; then
        PROFILE="default"
        current="1"
        index=$i
        break
    elif [[ "${PROFILES[$i]}" == "solar-only" ]]; then
        PROFILE="solar-only"
        current="2"
        index=$i
        break
    fi
done

echo "Select Powerwall-Dashboard Setup Profile"
echo ""
echo "Current Setup Profile: ${PROFILE}"
echo ""
echo "1 - default     (Powerwall w/ Gateway on LAN)"
echo "2 - solar-only  (No Gateway - data retrieved from Tesla Cloud)"
echo ""
while [ 1 ]; do
    read -p "Select profile: [${current}] " response
    if [ -z "${response}" ]; then
        response="${current}"
    fi
    if [[ "${response}" == "1" ]]; then
        if [[ "${PROFILE}" != "default" ]]; then
            changed=1
        fi
        PROFILE="default"
    elif [[ "${response}" == "2" ]]; then
        if [[ "${PROFILE}" != "solar-only" ]]; then
            changed=1
        fi
        PROFILE="solar-only"
    else
        continue
    fi
    break
done

# Update list of docker compose profiles and save to env file
if [ $changed -eq 1 ]; then
    if [ $index -ge 0 ]; then
        PROFILES[$index]="${PROFILE}"
    else
        PROFILES+=("${PROFILE}")
    fi
    sed -i.bak -E "/(^#|^)COMPOSE_PROFILES=/d" "${COMPOSE_ENV_FILE}"
    echo "COMPOSE_PROFILES=$(IFS=,; echo "${PROFILES[*]}")" >> "${COMPOSE_ENV_FILE}"
fi
```

This was before the Docker Compose Helper Functions so would need some further changes if we were to use it. But that was the gist of what I had in mind originally.

However I do think the current implementation is best to keep it simple.

mcbirse:
Looking into this issue further, Docker Compose V1 does actually support Profiles, **as long as** the version being used is `1.28.0` or newer...

I wonder if it would be better to check the version of Docker Compose before resorting to using the fix?

Or, perhaps, we could run something similar to below to check for errors instead, which validates the compose config.

```bash
if docker compose -f powerwall.yml $pwextend config > /dev/null 2>&1; then
    #all good
else
    #failed - assume no profiles support
fi
```

I do also wonder how many people are still using Docker version older than `1.28.0` (i.e. pre 2021-01-20) and should it be supported?

jasonacox:
Good call/fix on testing for docker V2 first. 

> I do also wonder how many people are still using Docker version older than 1.28.0 (i.e. pre 2021-01-20) and should it be supported?

Fair point! I happen to still have an old Ubuntu 18.04.6 system running and it has docker-compose 1.17.1.  I have intended to upgrade it but also seemed to find it mirroring some of the odd behaviors Synology NAS users would report.  This is the curse of any project. When do you stop supporting old version when it has spread to so many platforms? :) 

I ran this:

```bash
if docker-compose -f powerwall.yml config > /dev/null 2>&1; then
   echo "You have a current docker-compose"
else
   echo "You have an old docker-compose"
fi
```

It works.  "You have an old docker-compose".  I suggest we do something like this:

```bash
echo "Running Docker Compose..."
if docker compose version > /dev/null 2>&1; then
    # Build Docker (v2)
    docker compose -f powerwall.yml $pwextend $@
else
        # Build Docker (v1)
        if docker-compose -f powerwall.yml config > /dev/null 2>&1; then
            pwconfig="powerwall.yml"
        else
            echo "** WARNING **"
            echo "    You have an old version of docker-compose that will"
            echo "    be depreciated in a future release. Please upgrade or"
            echo "    report your use case to the project."
            echo ""
            echo "Applying workaround for old docker-compose..."
            pwconfig="powerwall-v1.yml"
            if get_profile "solar-only"; then
                pwconfig="powerwall-v1-solar.yml"
            fi
        fi
        docker-compose -f $pwconfig $pwextend $@
    else
        echo "ERROR: docker-compose/docker compose is not available or not running."
        echo "This script requires docker-compose or docker compose."
        echo "Please install and try again."
        exit 1
    fi
fi
```

This will still apply the workaround for now, but provide a user WARNING and hopefully prompt the community to report any edge use cases we need to consider before removing old docker-compose support.

What do you think?

setup.sh
jasonacox: I would say that over half of the install related issues were related to this. This is gold! Thanks.

setup.sh
jasonacox: On the default:

After I made that comment, in one of my test, I had hit enter twice after running the setup.  So your point is 100% valid. Let's keep it mandatory to enforce the user attention.

On future setup runs:

I'm fine with how it is for now.  However, the intent was that setup.sh could be run to fix a broken setup (user or other error).  For a future enhancement, it may be good to have setup.sh check compose.env and report what was selected and offer resetting.  Something like: "Existing configuration found: Your profile is set up for solar-only, do you wish to change that?"  If they do, nuke the COMPOSE_PROFILES setting and let setup.sh continue with setup (prompting for 1/2).

jasonacox:
I added these changes to a v3.0.2 branch - https://github.com/jasonacox/Powerwall-Dashboard/commit/5ed0838d0259f4df30b01aeb06b48506659729cb

mcbirse:
> What do you think?

> I added these changes to a v3.0.2 branch

Great solution, I like this!

I committed a small change... since the `compose.env` file is now defined in the globals as a variable, for consistency I replaced the hardcoded references with the variable. And also fixed a typo. 😉 

Looks good to merge to me.

jasonacox:
Thanks, @mcbirse !  Merged. v3.0.2 is live.

PR #394, opened 2023-11-24
Change files to run a two-gateway system

gallimorej:
This makes changes all over the place to run a two-gateway system. It essentially runs all the services side by side with the exception of grafana.


jasonacox:
Wow, nice!  Do you ahve it working?

gallimorej:
I do have it working! It's an inelegant solution at the moment. I basically took all the scripts and config and duplicated them for a "left" and a "right" system. The only one that I didn't do that with is the grafana service -- but I do have a left and a right dashboard.

A better solution would probably be to enable passing in a variable as a name for the system to be appended as a suffix to the running services. 🤔 That would eliminate a lot of the duplication.

jasonacox:
Great job @gallimorej !  I agree. That's worth investigating.

Issue #400, opened 2023-12-11
3.0.3 upgrade errors - MacOS


jgleigh:
compose.env: line 35: weather411,default: command not found

Looks like the setup.sh script is corrupting the compose.env file if there are already profiles set.

Original file: COMPOSE_PROFILES=default,weather411

Corrupted file after running upgrade: COMPOSE_PROFILES=default weather411,default

Running upgrade with no compose.env file works correctly.

jgleigh:
MacOS 14.1.2 in case there are weird difference between OSes.

jasonacox:
Thanks, @jgleigh  - I'll investigate. I thought I had tested an upgrade eon MacOS but likely forgot the weather411 path. @mcbirse If you have a chance, take a look as well in case I'm missing something simple.

mcbirse:
Hi @jasonacox - I'm not sure, but it seems it could be a compatibility issue with the separator (IFS) being set to a comma to process the bash array and update COMPOSE_PROFILES.

From what @jgleigh has shown, it looks like IFS does not get set to a comma so it remained as the default (space) when the compose.env file was updated. Must only affect some specific OS's or bash shells, as I tested many upgrade scenarios (but not on MacOS).

At a guess, the below might fix the issue. Have not tested yet but I'll try to do that tonight if I get time.

i.e. replace all occurrences of `IFS=,;` with `IFS=','` in all scripts - example of updated helper function below.

```bash
add_profile() {
    # Create default docker compose env file if needed.
    if [ ! -f ${COMPOSE_ENV_FILE} ]; then
        cp "${COMPOSE_ENV_FILE}.sample" "${COMPOSE_ENV_FILE}"
    fi
    if ! get_profile "${1}"; then
        # Add profile to COMPOSE_PROFILES and save to env file
        PROFILES+=("${1}")
        if [ -z "${COMPOSE_PROFILES}" ]; then
            if grep -q "^#COMPOSE_PROFILES=" "${COMPOSE_ENV_FILE}"; then
                sed -i.bak "s@^#COMPOSE_PROFILES=.*@COMPOSE_PROFILES=$(IFS=',' echo "${PROFILES[*]}")@g" "${COMPOSE_ENV_FILE}"
            else
                echo -e "\nCOMPOSE_PROFILES=$(IFS=',' echo "${PROFILES[*]}")" >> "${COMPOSE_ENV_FILE}"
            fi
        else
[...]

PR #401, opened 2023-12-13
V3.0.4 - Fix Upgrade for MacOS

jasonacox:
Fix `upgrade.sh` to address issue where get_profile() function incorrectly reads compose.env causing upgrade failure. Closes issue #400.


mcbirse:
Hi Jason, but I think this fix should have been applied to the other scripts as well.

The same line is used in 4 other scripts and were not updated (`compose-dash.sh`, `setup.sh`, `verify.sh`, and `weather.sh`).

I pushed the same change to those scripts - hope that is okay to do.

i.e. from:
```
IFS=',' read -a PROFILES <<< ${COMPOSE_PROFILES}
```
to:
```
IFS=',' read -ra PROFILES <<< "${COMPOSE_PROFILES}"
```

jasonacox:
Yes, thank you!!!

PR #410, opened 2023-12-20
Update influxdb.conf

youzer-name:
Change monitor: store-enabled to false to reduce CPU load


jasonacox:
Thanks @youzer-name ! 🙏

PR #414, opened 2023-12-27
v4.0.0 - Add Cloud Mode support via pyPowerwall v0.7.0

jasonacox:
This update will use the cloud mode support built in to pyPowerwall v0.7.0 to pull live system metrics from the Tesla Owners API. 

Related
* Better support for Tesla Solar-only and Powerwall 3 owners that don't have a local energy gateway API - See #387 
* Transition update in case other APIs are removed in the future - See #402 

Updates
* Using pypowerwall for both local and cloud mode (replaces Tesla-history service, but the Tesla-history tool will continue to be used to fill in historic data or gaps) - See related https://github.com/jasonacox/pypowerwall/pull/59
* Removal of Docker Compose profiles (helps with some older systems that don't fully support this) and the v1 related legacy support.
* Updated `setup.sh` and `upgrade.sh`  to support transition to pyPowerwall for cloud mode.
* Updated `verify.sh` to support cloud mode metrics.
* Updated dashboard for solar-only users to include power flow animation panel.


youzer-name:
Is this update going to push my edge case right off the edge of the project? (When I just got back onboard days ago!). 

I need to run tesla-history to pull data from my solar site which is separate from my Powerwall site. Should I deploy two completely separate Powerwall Dashboard environments, one set up for solar only and one for Powerwall?

jasonacox:
> I need to run tesla-history to pull data from my solar site which is separate from my Powerwall site. Should I deploy two completely separate Powerwall Dashboard environments, one set up for solar only and one for Powerwall?

That would work.  The pyPowerwall proxy will be able to pull from either site (it is selectable) and you could technically run two instances, one for solar one for PW.  But if you could wave a magic wand, what would be the idea outcome/solution?  This may be a good time to see if we can accommodate.

youzer-name:
With the current setup it's really simple to run both. I used the powerwall.extend.yml file to move tesla-history from the solar-only profile to the default profile, so it runs alongside the local powerwall setup.  

I'm not 100% clear on how the new local/cloud hybrid is going to work, but the main thing I'd do with my magic wand is make it so I can (1) point the Tesla History container to my solar PPA site and (2) have the configuration persist through running the upgrade.sh script.

mcbirse:
Hi @youzer-name - I believe I recall you have the tesla-history container also pulling data into a different DB in InfluxDB?

One important distinction here is with the pypowerwall setup, data is being retrieved by telegraf and output to InfluxDB raw database and then downsampled by CQ's. Running 2x pypowerwall would be problematic.

To have pypowerwall running for both a Powerwall site and a Solar-Only site but with data going to different databases, this would mean another separate set up for telegraf, raw database and InfluxDB CQ's I think... sounds non-trivial.

Your current set up should be able to be maintained I believe (i.e. using the tesla-history container) even after these latest changes, if you keep using the powerwall.extend.yml setup and pypowerwall exactly as you have been.

> (2) have the configuration persist through running the upgrade.sh script.

Not sure if you noticed, but with the latest change I made you should be able to have you config persist now since you can put your custom InfluxDB config settings into influxdb.env, or is there something still preventing this persistence when upgrading?

youzer-name:
Actually, while your recent change will probably end up useful later, the specific change in InfluxDB that I was trying to make became part of the base project configuration... that was to turn off internal logging out reduce InfluxDB CPU usage. So for the moment I don't need to customize InfluxDB.  I'm pretty sure the database config for Solar-Only was already persisting across upgrades since everything is set up in Powerwall-Dashboard/tools/tesla-history/tesla-history.conf and that's where I have it pointed to my 'solar-only' database.

As long as there is still a way to have the Tesla-History container run on a periodic schedule, as opposed to only running on demand, I should be ok.

jasonacox:
I toggled this PR to ready to start the work, but there is more to do. @mcbirse would love your thoughts on the best way to treat the "cloud mode" going forward. Feel free to commit to the v4.0.0 branch directly.

So far, I updated the `setup.sh` and `upgrade.sh` to create the cloud mode pypowerwall files (`.pypowerwall.auth` and `pypowerwall.site`) for the updated volume binding in Docker compose. I staged version v4.0.0 with [pypowerwall v0.7.1.](https://github.com/jasonacox/pypowerwall/releases/tag/v0.7.1) More to do...

The new pypowerwall proxy (t35) version will give users the ability to get live data, including the power flow animation, for systems that don't have a local API . This means the pypowerwall container (`jasonacox/pypowerwall:0.7.1t35`) will support traditional Powerwall owners as well as Solar-only and new Powerwall 3 owners (as mentioned in #387 ).  Anyone wanting to manually add/keep tesla-history can do so in the `powerwall-extend.yml` as @youzer-name mentions. And of course, the standalone [tesla-history tool](https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools/tesla-history) can still be used to fill in historic data as well as any gaps that form over time.

But, I'm wondering about the best way to handle the `upgrade.sh` migration for other existing users. Some initial thoughts:

1. Keep the docker-compose profiles and allow users to continue with their tesla-history setup and just add a third setup option (pypowerwall cloud-mode). Users wanting to switch would re-run setup, resulting in tesla-history being removed and pypowerwall being added. Upside is that this could be less disruptive for edge case users. Downside is that this would translate to tech-debt of managing the profiles and 3 different options going forward.
2. Remove the docker-compose profiles and replace tesla-history setup with pypowerwall cloud-mode. Upside is that this is simple/elegant, easier to maintain and less confusing to users (2 modes: local or cloud). Downside is that it will take a bit to figure out how to unwind the profiles and "clean up" existing installations without breaking things. 
3. Others?

I'm leaning toward 2 as it seems like the best long term solution, even if a bit more upfront work to make it happen. What am I missing?

mcbirse:
> I toggled this PR to ready to start the work, but there is more to do. @mcbirse would love your thoughts on the best way to treat the "cloud mode" going forward. And please, free to commit to the v4.0.0 branch as well.

No problem - I will have a look at what changes we need to make.

> So far, I updated the `setup.sh` and `upgrade.sh` to create the cloud mode pypowerwall files (`.pypowerwall.auth` and `pypowerwall.site`) for the updated volume binding in Docker compose. I staged version v4.0.0 with [pypowerwall v0.7.1.](https://github.com/jasonacox/pypowerwall/releases/tag/v0.7.1) More to do...

I actually had a thought about this one yesterday already.... I was thinking, it may be better to add a new folder in the repository called e.g. `.auth` (and add a `.keep` file inside, like the Grafana folder has), then the Docker Compose volume binding should be to the `.auth` folder (instead of the individual files). Cloning the repository will include the folder so the container will always start, even if the .pypowerwall.auth/.site files within are missing. No requirement to create the files in setup/upgrade scripts, and they will be created by pypowerwall during setup.

> But, I'm wondering about the best way to handle the `upgrade.sh` migration for other existing users.

My preference would be for option 2.

I think for the solar-only users we should be able to provide some instructions for using the tesla-history script as a container, e.g. explaining and providing an example of what needs to be added to a `powerwall.extend.yml` for instance.

jasonacox:
> I was thinking, it may be better to add a new folder in the repository called e.g. .auth

I agree. I'll update pypowerwall proxy to support passing an PW_AUTH_PATH variable.

mcbirse:
> I agree. I'll update pypowerwall proxy to support passing an PW_AUTH_PATH variable.

Ahh, that's not what I meant, sorry. I don't think any change to pypowerwall is required. Just that the volume bind would be to a folder instead of the files.

jasonacox:
Ha!  Ok, let me know if you can figure it out. That would be great. I staged a new v0.7.2 branch just in case: https://github.com/jasonacox/pypowerwall/commit/743591be44dcf9981aa61ba10f7ce074cd3e2072

mcbirse:
> I'll update pypowerwall proxy to support passing an PW_AUTH_PATH variable.

Sorry, yes you are correct, we would also need a variable added to pypowerwall to set the path to where the auth files are so it points to the target folder of the compose config. Overall I think this will be a better setup in the long wrong.

jasonacox:
I updated to pypowerwall (jasonacox/pypowerwall:0.7.2t36) to now allow an environmental setting for the location of the .pypowerwall.auth and .pypowerwall.site files.  It is set with `PW_AUTH_PATH`.

I think steps are:
* Create an `.auth` folder with a `.keep` file.
* Add `PW_AUTH_PATH=.auth` to the pypowerwall.env.sample
* For upgrade, we will need to scan the local `pypowerwall.env` to see if the PW_AUTH_PATH exists, and if not, add it to be `PW_AUTH_PATH=.auth`

jasonacox:
> I think for the solar-only users we should be able to provide some instructions for using the tesla-history script as a container, e.g. explaining and providing an example of what needs to be added to a powerwall.extend.yml for instance.

I'm probably missing something obvious. I assumed that Tesla solar-only customers can still use the Tesla app to see there current production and grid usage which would mean that we should be able to use the same Tesla Owners API to get that data.  I followed your Tesla-history script to grab both battery as well as solar lists in pypowerwall:

```python
sitelist = self.tesla.battery_list() + self.tesla.solar_list()
```

Should all the solar-only data also show up in api/1/energy_sites/{site_id}/live_status ?  If so, if we solve for the "cloud mode" in setup.sh and upgrade.sh, solar-only customers should be able to start using the same dashboard and setup as Powerwall owners. I think?

mcbirse:
> I'm probably missing something obvious.

Sorry, no - I think I've had too many late nights and I did not explain properly! What I meant was for the edge cases only - like @youzer-name 

> Should all the solar-only data also show up in api/1/energy_sites/{site_id}/live_status ? If so, if we solve for the "cloud mode" in setup.sh and upgrade.sh, solar-only customers should be able to start using the same dashboard and setup as Powerwall owners. I think?

Yes, I would say the majority of solar-only users can use the new cloud mode of pypowerwall.

> * For upgrade, we will need to scan the local `pypowerwall.env` to see if the PW_AUTH_PATH exists, and if not, add it to be `PW_AUTH_PATH=.auth` (TODO)

Working on this - I have a slightly different approach which may be more sensible. I will commit changes but feel free to choose your preference though.

jasonacox:
Thanks @mcbirse! Can't wait to see your approach! Please continue.

PS - Happy New Year!! 🥳

mcbirse:
Thanks! Happy New Year! (in advance for your tomorrow) 🥳

mcbirse:
I made changes to pypowerwall & proxy as the cloud setup mode was not working when running as a container. i.e. `docker exec -it pypowerwall python -m pypowerwall setup` which will be needed during the setup script.

Have submitted a PR for that. I had also noticed a delay in the container stopping/restarting which the change to the signal handler should fix (untested).

For the PW_AUTH_PATH environment, I have committed changes around what I was thinking may be better.

i.e. the environment variable is defined in the Docker Compose config only - not pypowerwall.env. It seems clearer, as it is really only relevant to the compose config and then easily identifiable in the file along with the volume bind mount source/targets which it relates to.

Also, if it was defined in pypowerwall.env file, a user looking at that may think they could change the path defined here to point their auth files at a different folder.... but that won't work and will confuse people, because that path is for within the container environment, not the local filesystem.

The paths could be changed/overridden by a user redefining those attributes in the powerwall.extend.yml file instead.

My thoughts/rant on this. Happy for changes if you see a better way or prefer the previous setup. I also haven't tested all this yet since it needs updates for pypowerwall & proxy to be pushed.

Taking a break for NYE now. 😄

mcbirse:
Per the changes from https://github.com/jasonacox/pypowerwall/pull/62 I have tested the cloud mode setup, and running `docker exec -it pypowerwall python -m pypowerwall setup` works... Now to incorporate that into the setup scripts.

I've made a start on this (with the intention to unwind the use of Docker Compose profiles for better compatibility).

More work to do yet, but doable... Hardest part will probably be ensuring the upgrade process will work for all of the various installation iterations out there.

jasonacox:
Thanks @mcbirse !  Please feel free to make / commit any changes when you have time. 

I'm going to work on adding the CQs for the new data we are gathering with pypowerwall v0.7.x, specifically the aggregate Powerwall capacity data. This will give those who don't have vitals (or lost vitals) that important data in the dashboard.  It was a top data point the community wanted to keep.

image

We also added `time_remaining_hours` data. I a playing with that data in the dashboard either as a meter value or possibly as a component of the Powerflow Animation.

image

image

I'm not sure yet what would look best or if there is any interest in that data on the dashboard, but it will be available.

youzer-name:
@jasonacox  Are you calculating the battery time available in realtime based on the current values? If so, I'm not sure how useful a metric that will really be. An hour before the sun goes down it will produce an overly optimistic number. An hour before sunrise, an overly pessimistic number. 

I have two panels in my 'experimental' dashboard where I played around with this. One of them is based on the last 24 hours... ( battery capacity + last 24 hours solar production) / (last 24 hours home use) = hours remaining. I actually am using the last 24 hours of home use not including car charging, as I wouldn't do any car charging if the power was out. 

image


I have another panel that is based on the last hour of home use and doesn't account for car charging, and since one of the cars is drawing ~8 kW right now, it says I'd only have 3 hours of battery time!

image

jasonacox:
Thanks @youzer-name !  That helps confirm that it isn't valuable. I'm computing it but also able to get that number from Tesla Cloud (the same number will show up in the Tesla app) and it is equally useful (or useless 🤷 ).  😁 

I'll focus on the Powerwall Capacity aggregate for this release. I know that will be useful for Powerwall 3 owners that don't have access to any local data that includes the separate Powerwall capacity metrics.

jasonacox:
For anyone needing (or wanting) to test cloud mode:

You can manually activate the cloud mode in your Powerwall-Dashboard setup by doing this:

1. Backup your config (just to be safe) 
2. From `Powerwall-Dashboard` directory run: `mkdir .auth`
3. Edit the `powerwalls.yml` file and replace the pypowerwall section with this:
```yml
    pypowerwall:
        image: jasonacox/pypowerwall:0.7.3t37
        container_name: pypowerwall
        hostname: pypowerwall
        restart: unless-stopped
        volumes:
            - type: bind
              source: .auth
              target: /app/.auth
        user: "${PWD_USER:-1000:1000}"
        ports:
            - "${PYPOWERWALL_PORTS:-8675:8675}"
        environment:
            - PW_AUTH_PATH=.auth
        env_file:
            - pypowerwall.env
        profiles:
            - default
```
4. Edit the `pypowerwall.env` file and remove the IP address of the Powerwall Gateway (`PW_HOST=`)
5. Run `./compose-dash.sh up -d` and it will download the latest pypowerwall and switch to cloud mode. 
6. Before it will work, you need to set up the Tesla auth:  `docker exec -it pypowerwall python -m pypowerwall setup`
7. Restart: `docker restart pypowerwall`

mcbirse:
@jasonacox - I'll have some further updates soon (weekend perhaps). Managed to find a little time last night to work on the setup script some more.

However, this resulted in likely another update to pypowerwall I think, and weather411 (small bug fix), which will need push to PyPI / Docker Hub etc. again.

I'd much rather be spending my time on this project than my real job! 😄 C'est la vie.

jasonacox:
@mcbirse No worries and no rush! And, I'm right there with you! Been crazy at work too... welcome to 2024. 😁

Thanks for all your help! 🙏

mcbirse:
Latest commits are work in progress, and require more testing.... new container versions of pypowerwall and weather411 will need to be built and pushed for the Docker Compose changes before being able test.

TODO: Upgrade script, verify script, more testing... and anything else overlooked?

jasonacox:
I accepted the pypowerwall change and pushed the new v0.7.5 version to PyPI and proxy to dockerhub.

billraff:
> 6. Before it will work, you need to set up the Tesla auth:  `docker exec -it pypowerwall python -m pypowerwall setup`
> 7. Restart: `docker restart pypowerwall`

Hi Jason

Just following along to test the Cloud version and I am at step 6. When I issue that command the response is -

billr@billrs-Mac-mini Powerwall-Dashboard % docker exec -it pypowerwall python -m pypowerwall setup
pyPowerwall [0.6.4]

Usage:

    python -m pypowerwall [command] [<timeout>] [-nocolor] [-h]

      command = scan        Scan local network for Powerwall gateway.
      timeout               Seconds to wait per host [Default=1.0]
      -nocolor              Disable color text output.
      -h                    Show usage.

and nothing else happens.

jasonacox:
Hi @billraff  - I see the problem:

> pyPowerwall [0.6.4]

You need to upgrade to at least version 0.7.3 which is what step 3 should have done.  Did you make that edit?  Your install may be using the old version (check `powerwall-v1.yml`).  In any case, there is a newer version with some other updates now too, change those referenced yml files to `jasonacox/pypowerwall:0.7.6t39` - you can also manually remove the old images:

```bash
#
# stop and remove old one
docker stop pypowerwall
docker rm pypowerwall

# remove all old images
docker images | grep pypowerwall | awk '{print $3}' | xargs docker rmi -f

# start the stack
./compose-dash.sh up -d
```

upgrade.sh
```diff
@@ -183,14 +139,32 @@ if [ -f tools/solar-only/compose.env ] && [ ! -f ${COMPOSE_ENV_FILE} ]; then
...
+if [ "${PROFILE}" == "solar-only" ]; then
+    # Create Powerwall Settings
+    rm -f ${PW_ENV_FILE}
+    mkdir -p .auth
+    cp tools/tesla-history/tesla-history.auth .auth/.pypowerwall.auth
+    if grep -qE "^SITE = [0-9].+" tools/tesla-history/tesla-history.conf; then
+        grep -E "^SITE = [0-9].+" tools/tesla-history/tesla-history.conf | awk '{ print $NF }' > .auth/.pypowerwall.site
     fi
-    echo ""
+    EMAIL=$(grep -E "^USER = .+@.+" tools/tesla-history/tesla-history.conf | awk '{ print $NF }')
```
jasonacox, line 154: This migration logic is clever! ! I'll run it across my various OS systems but seems like it would be universal.

weather/server.py
jasonacox, line 1: Great additions here! I pushed this version to DockerHub and tested it.

jasonacox/weather411:0.2.3 ✅

setup.sh
```diff
@@ -338,6 +287,35 @@ fi
 ./compose-dash.sh up -d
 echo "-----------------------------------------"
 
+# Run Local Access mode network scan
+if [ "${config}" == "Local Access" ] && ! grep -qE "^PW_HOST=.+" "${PW_ENV_FILE}"; then
+    echo "Running network scan... (press Ctrl-C to interrupt)"
+    if type winpty > /dev/null 2>&1; then
+        IP=$(netstat -rn | grep "0.0.0.0" | awk '{ print $4 }' | head -1)
+    else
+        IP=$(ip route get 8.8.8.8 | awk '{ print $7 }')
```
jasonacox, line 296: This doesn't work on my Mac.  I can use `ipconfig getifaddr en0`  or `ifconfig | grep 'inet ' | grep -v '127.0.0.1' | awk '{print $2}'`.

maybe something like this?

```bash
OS=$(uname -s)

# Get local IP based on the operating system
case $OS in
  Linux*) IP=$(ip route get 8.8.8.8 | awk '{ print $7 }') ;;
  Darwin*) IP=$(ifconfig | grep 'inet ' | grep -v '127.0.0.1' | awk '{print $2}') ;;
  CYGWIN*|MINGW32*|MSYS*) IP=$(netstat -rn | grep "0.0.0.0" | awk '{ print $4 }' | head -1) ;;
  *) IP=""
esac

echo "Your OS is ${OS} and your local IP is ${IP}"
```

mcbirse: That looks better than what I did. I will incorporate this or similar into the script - feel free to change it though.

I was trying to find a consistent OS independent way of discovering the local IP address in bash.... it seems there isn't one?

jasonacox: Sounds good!  Yes, search for one and even asked ChatGPT! 😂  No luck. Worse case, if it gets an empty or wrong, response, the scan won't rung and the user can manually enter the IP during setup.

mcbirse: Yep true. When you think something should be simple... lol

To complicate things, small change for this I'll remove the "32" in `CYGWIN*|MINGW32*|MSYS*)` as my Git Bash returns MINGW64 😆

jasonacox: It looks good and we are getting close @mcbirse. What left on your list? I'll start testing.

mcbirse:
> It looks good and we are getting close @mcbirse. What left on your list? I'll start testing.

@jasonacox - I think we are at "almost ready for release" stage, pending some further testing and updates to documentation / instructions / release notes, etc.

I have tested the following successfully (both Linux and Windows, but not Mac):

- Fresh install of v4.0.0 by running `setup.sh` after cloning repository (both Local Access mode and Tesla Cloud)
- Switching config modes from/to Local Access and Tesla Cloud by re-running `setup.sh`
- Upgrade from v2.10.0 to v4.0.0 (Local Access mode)
- Upgrade from v2.10.0 to v4.0.0 (beta Solar Only install pre- compose-profiles, to test migration)
- Upgrade from v3.0.5 to v4.0.0 (Local Access mode)
- Upgrade from v3.0.5 to v4.0.0 (Solar Only mode, post- compose-profiles to test migration)
- The `verify.sh` script in various states from good to things being broken
- ...and probably some other scenarios that I now can't recall!

mcbirse:
For anyone else that would like to test and provide feedback, the manual instructions here https://github.com/jasonacox/Powerwall-Dashboard/pull/414#issuecomment-1882355993 can be followed for an existing install.

Or, to test a clean install:

```bash
#Clone v4.0.0 branch
git clone -b v4.0.0 https://github.com/jasonacox/Powerwall-Dashboard.git
```

```bash
#Install Powerwall-Dashboard v4.0.0
./setup.sh
```

jasonacox:
## v4.0.0 on RPi/Linux:
* Upgrade ✅  - I have to figure out a better way to test. For now I just `git checkout v4.0.0; cp upgrade.sh tmp.sh; bash tmp.sh upgrade` which ran clean (`./verify.sh`):
* New Install ✅  - The "scan for IP" feature is nice but slow. I wonder if we should somehow hint that entering the IP if you know it is the recommended path. Maybe something like?
```
   IP Address (recommended or leave blank to scan network): 
      or
   IP Address (leave blank to scan if unknown): 
```

## v4.0.0 on MacOS
* Upgrade ✅ 
 ❌ One  Issue - This is due to my method of testing upgrade (checkout brings over the powerwall.yml from the new version first so when it runs to shutdown the old containers, it is missing the influxdb.env file and fails). Fixed with manual `cp influxdb.env.sample influxdb.env` then it ran correctly.
* New Install ✅
🟡 Setup with latest Docker Desktop trips this warning: WARNING: You do not appear to be in the docker group. No problems proceeding.

## v4.0.0 on Windows
* Upgrade ❔ 
* New Install ❔ 
I still need to test Windows... 

## Verify

image

larosseqnx:
Upgraded to most recent using upgrade.sh.

Followed the above manual cloud transition steps and pypowerwall is no longer acquiring data.  After running docker exec -it pypowerwall python -m pypowerwall setup I paste in teh requested URL and it fails with.

------------------------------------------------------------
Missing auth file .pypowerwall.auth - run setup

ERROR: Failed to connect to Tesla Cloud
ERROR: Failed to setup Tesla Cloud Mode

larosseqnx:
Verify Powerwall-Dashboard 3.0.8 on Linux - Timezone: America/Los_Angeles


Checking configuration
 - Dashboard configuration: default
 - EnvVar COMPOSE_PROFILES: default,weather411

Checking pypowerwall
 - Service (port 8675): ERROR: Not Listening
 - Version: Unknown
 - Powerwall State: ERROR: Not Connected - Firmware Version: Unknown

Checking telegraf
 - Container (telegraf): GOOD

Checking influxdb
 - Environment File influxdb.env: GOOD
 - Container (influxdb): GOOD
 - Service (port 8086): GOOD

Checking grafana
 - Container (grafana): GOOD
 - Service (port 9000): GOOD

Checking tesla-history
 - Skipped: Only required in 'solar-only' configuration

Checking weather411
 - Config File weather/weather411.conf: GOOD
 - Service (port 8676): GOOD
 - Weather: {"temperature": 58.1}
 - Version: 0.2.2

larosseqnx:
Followed upgrade procedure and reran setup.

@teslacharts Powerwall-Dashboard]$ ./verify.sh 
Verify Powerwall-Dashboard 4.0.0 on Linux - Timezone: America/Los_Angeles


Checking pypowerwall
 - Service (port 8675): GOOD
 - Version: 0.7.6 Proxy t39
 - Powerwall State: CONNECTED - Firmware Version: 23.44.0
 - Cloud Mode: YES - Site ID: 2252153317443337 (Hoffman St)

Checking telegraf
 - Container (telegraf): GOOD

Checking influxdb
 - Environment File influxdb.env: GOOD
 - Container (influxdb): GOOD
 - Service (port 8086): GOOD

Checking grafana
 - Container (grafana): GOOD
 - Service (port 9000): GOOD

Checking weather411
 - Service (port 8676): GOOD
 - Weather: {"temperature": 58.32}
 - Version: 0.2.3

All tests succeeded.

larosseqnx:
Animation working and im seeing data in the alerts field.  Granted only a single alert "System Connected to Grid"
image

jasonacox:
> Animation working and im seeing data in the alerts field. Granted only a single alert "System Connected to Grid"

Thanks for confirming the animation works! 🚀  

As to the alert, yes, without the vitals payload, we are limited. I plan to spend some time to figure out if we can infer other alerts based on various signals in the payloads. We might be able to add some new ones that would be helpful.

mcbirse:
> * The "scan for IP" feature is nice but slow. I wonder if we should somehow hint that entering the IP if you know it is the recommended path.

Implemented multi-threaded network scanning in latest pypowerwall to scan multiple hosts simultaneously... this should increase the network scan speed significantly.

jasonacox:
I'm running through the tests again but this is looking ready to merge.

mcbirse:
I can't recall the status of this, but should `dashboard-solar-only.json` be updated to include the Powerflow meter again?

jasonacox:
Great idea!  I added it back... I should probably look at a way to modify the display to remove the Powerwall icon, but that can be a future edit (or someone else can help figure that out :).

image

jasonacox:
I'm going to do a quick pass through the documentation and then merge. 🚀

jasonacox:
Thanks to everyone who tested for us!   If you switched to the v4.0.0 branch, you will want to switch back to main and upgrade:

```bash
git checkout main
./upgrade.sh
```

Thanks for all the work on this @mcbirse ! 🙏

BJReplay:
Nice work @mcbirse and @jasonacox.  I just upgraded and it nicely handled my extend.yml for my ecowitt weather service without a hiccup.

I've just got my control service using the fleet API for setting modes (self-consumption, autonomous for fast charging, backup for slow charging) and backup percentages, and it seems to be working just as well as the owner API used to work.

Now that I've got it working on the fleet API, I think a re-write from VB.Net to Python is next cab off the rank (which means I could submit a massive PR for control to this repo), but that is a long way down the track, I fear.

jasonacox:
Thanks @BJReplay !

On the FleetAPI, I had started on the python version here: https://github.com/jasonacox/pypowerwall/tree/main/tools/fleetapi - My TODO is to add this into pypowerwall's cloud class to allow user to pick FleetAPI (if they sign up for it) or the Tesla Owners API from TeslaPy. We could also set it up so that it could be used for other tools/projects (just `from pypowerwall import fleetapi` or something) which would expose easy to use functions for making FleetAPI calls.

BJReplay:
> On the FleetAPI, I had started on the python version here:

I wish I'd found your version before I muddled my way through (using Postman).  I got to POST /api/1/partner_accounts pretty quickly, but every call I made got 412 Precondition Failed with body
```json
{
"error":"Account  must be registered in the current region https://fleet-api.prd.na.vn.cloud.tesla.com/, please see https://developer.tesla.com/docs/fleet-api#register"
}
```

which, of course, was exactly what I was trying to do - register :)

I gave up for a couple of days, came back, and tried again, and it worked first time, so I put it down to guessing that maybe their service wasn't working the first time I tried, or it takes some time between creating your public key and being able to register, but from your readme, it looks like once you've registered, and generated your key, it should be pretty much automatic?

jasonacox:
>  it looks like once you've registered, and generated your key, it should be pretty much automatic?

That's correct. Once registered you use the Client ID and the one-time-code to generate your auth token and refresh token. The sticky bits are that it requires you to have your own website (domain with a PEM file stored on it for registration) and sign up with a business name which may not work for everyone. Also, there are rate limits which don't seem like a problem for those of us using for a hobby, but could be an issue if you are signing up others under your Client ID. Good news is that it is free for now and it is an official API from Tesla vs the unofficial version we have been using. 😁

PR #426, opened 2024-01-28
v4.0.1 - Powerwall Voltages

mcbirse:
## v4.0.1 - Powerwall Voltages

* Add CQs and update `dashboard.json` to include voltages from battery block vitals. This addresses users no longer seeing Powerwall voltages in the dashboard with firmware 23.44.0 or later - See related https://github.com/jasonacox/Powerwall-Dashboard/discussions/109#discussioncomment-8263020


jasonacox:
This is great!  I'm seeing the data already.

In the US, the new data is is based on the combined L1/L2 lines (phase) so we have the old data tracking at 122V and the new data at 244V.

image

I created an Override using the regex `.*[ ](L[1-2])\b` for the name to identify the L1/L2 labels and push them to the right axis.

image

image

But I'm not sure it is worth adjusting as it probably wouldn't work outside the US and all of this L1/L2 Vsplit data will be gone after vitals are removed.

I'm in favor of proceeding with the merge. Everything else looks good.

jasonacox (approved):
Looks great @mcbirse !

mcbirse:
> But I'm not sure it is worth adjusting as it probably wouldn't work outside the US and all of this L1/L2 Vsplit data will be gone after vitals are removed.

Great solution with the regex based axis override! I just tested this and it seems acceptable for 220V-240V standards as well.

I think it is worth including in that case, so as not to mess up the graph for 120V users. I will include your change and merge this. It also highlighted an extraneous override "Solar Energy" (color scheme) that was probably copied by accident from another panel, which I will remove.

For 220V-240V the left/right axis may be slightly skewed, but it's pretty minor.

image

The oddity of the graph with both a left/right axis will eventually disappear as more people are upgraded to the new firmware and vitals are removed.

image

In the meantime, having the axis split seems sensible while there is a transition period and PW voltages are being displayed from both vitals and battery blocks at the same time.

jasonacox:
Excellent! Thanks @mcbirse 🙏

jasonacox:
Hi @billraff, I doubt this since you just did an upgrade, but did you happen to switch to `cloud mode`? Run `./verify.sh` or go to http://pypowerwall:8675/help to see what mode it is running.

Try going to  http://pypowerwall:8675/api/system_status - that is where the new voltage data is coming from. If that isn't working, there is something with InfluxDB not running the CQs, otherwise you would see it in the updated `dashboard.json` file that you loaded.

billraff:
Yes, this is on my test box so it is running cloud mode:

Verify Powerwall-Dashboard 4.0.1 on MacOS - Timezone: America/Los_Angeles


Checking pypowerwall
 - Service (port 8675): GOOD
 - Version: 0.7.7 Proxy t40
 - Powerwall State: CONNECTED - Firmware Version: 23.44.0
 - Cloud Mode: YES - Site ID: 2252262374693454 (Powerwall)

Checking telegraf
 - Container (telegraf): GOOD

Checking influxdb
 - Environment File influxdb.env: GOOD
 - Container (influxdb): GOOD
 - Service (port 8086): GOOD

Checking grafana
 - Container (grafana): GOOD
 - Service (port 9000): GOOD

Checking weather411
 - Service (port 8676): GOOD
 - Weather: {"temperature": 68.97}
 - Version: 0.2.3

Not sure where I should be looking for that data.

Screenshot 2024-01-29 at 1 17 05 PM

jasonacox:
> Cloud Mode: YES 

You are running in cloud mode. Unless you have a Powerwall 3, you should be able to switch to local mode to get the additional data. You can do that by running the `setup.sh` again and selecting `local mode`. 

Cloud mode doesn't provide voltage data. As you can see in the payload `"battery_blocks": []` is empty.

PR #430, opened 2024-01-31
Update dashboard.json

youzer-name:
Stack powerwall current energy values in Powerwall Capacity panel. Add "Off Grid" indicators to the bottom of the Frequency and Voltages panels. Set all 3 panels "Connect null values" = "Threshold 1h"


jasonacox:
I tested it and it looks good @youzer-name but I have a question. How can we align the Powerwall stacking axis to the same as the "fully charged" line? Having it on separate axis causes it to go above the "fully charged" lines. I tried myself, but haven't found it yet.

youzer-name:
If I understand correctly, you want the 'full pack' lines to stay above the stack of 'current energy' lines. I think if you remove the soft max on the main chart settings that might do it. With the soft min at zero, it should adjust the left axis to keep the values at the very top. This might not work for installations where multiple powerwalls have very different full pack capacities, as the lower-capacity packs will chart further from the top of the axis.

youzer-name:
What do you think of this?

image

This puts everything on one axis. The 'full' lines are placed so that each one shows the 'space' between that pack's current capacity and that pack's full capacity. The axis is not relevant to the placement of those lines, but they don't appear in the tooltip and are only visible in the legend.

Here is a view of both options together. The top one keeps the full rated capacity at the top using the left axis. The bottom one puts the full rated capacity relative to each PW's current charge.

image

I like this for a 24 hour view or multiple day view. This is probably not ideal for a multi-month view where you would want to see the change in the full rated capacity over time.

I still think the base project should consider splitting the existing dashboard into at least two dashboards, one for current status and one for historical data. The first would default to 'last 24 hours' and be optimized for viewing periods of hours or days, and ideally fit on most device screens without scrolling. The second would default to 'last 30 days' or 'last 90 days' and be optimized for viewing periods of months or years. The current status dashboard would be set for 30 second or 1 minute auto-refresh. The historical dashboard would have auto-refresh turned off or set to a much higher value since you really don't need 1-minute updates when looking at 90 days of data.

Relevant to this discussion, the current status dashboard would use the relative full values as per my latest creation. The rated full pack capacity isn't going to change meaningfully over a day or a week. The historical dashboard might show ONLY the full rated capacities as that is something worth tracking over long periods of time, but the daily charge/discharge state of the batteries might be less interesting.

That is the way my custom dashboards are set up. I don't show monthly totals and the annual production on my 'current status' dashboard, and I don't show the 'energy usage' panel on my historical dashboard. Nothing would prevent a user from setting the 'current status' to a 24 month view and getting this result in the 'energy usage' panel:

image

But it reminds me of a 1990's Windows screensaver and really doesn't convey a lot of useful information, plus I think my Raspberry Pi 4 just let off a puff of smoke when the dashboard loaded 😏

longzheng:
Agree with the "connect null values" change, however I think it should also be added to the "Energy Usage" graph.

For example during this day I lost connectivity to the Gateway but from the graph alone it's very hard to tell

image

But if I change "connect null values" to a threshold you can easily see it

image

jasonacox:
Thanks @longzheng, I made this update to the Energy Usage graph too.

Thanks @youzer-name !  I like these updates. Minor but effective. 

Releasing this as v4.0.2

youzer-name:
'Minor but effective"... I think you just came up with what they're going
to write on my tombstone.
😜

Thanks @youzer-name <https://github.com/youzer-name> ! I like these
> updates. Minor but effective.
>
> Releasing this as v4.0.2
>

mcbirse:
Thanks @youzer-name - these additions are great. Apologies I did not have a chance to review/test before the merge, have been unwell for a few weeks.

Nice epitaph, lol.

longzheng:
@youzer-name deserves all the credit, I didn't do any work 😇

Issue #431, opened 2024-01-31
Grid Status, Savings, Self-Powered, Solar Energy Year panels no longer work


sassyass:
These were all working fine last year, but about a month ago (the week before Christmas maybe?) these panels stopped working and are now spitting out errors.

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/15484580/f152a866-13a6-47dd-a30b-fa441017598b)
![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/15484580/bd4b8054-7f4a-4cb9-bf5b-8292d708eb94)

When I look at them they are all giving the same error: "Status: 500. Message: InfluxDB returned error: error parsing query: found \/, expected identifier, string, number, bool at line 1, char 149"

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/15484580/a37e692d-a906-442c-8bda-e09f88b4238a)

I took a peek with Chronograph, and it does look like values are still being populated into InfluxDB:
![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/15484580/b2a38004-d3e8-4e79-806d-41a78623ed02)

There was probably a pypowerwall update before this but not 100% sure; and I was waiting for more updates in case it would magically fix itself, but it's still persisted even after the latest updates.

Any help would be appreciated.

jasonacox:
Hi @sassyass - did you import the latest `dashboard.json` and overwrite the existing one?

sassyass:
I did import the latest `dashboard.json`, but it still gives the same errors.

jasonacox:
Yikes. This seems more like a corruption in the influxdb. Can you try these:

```bash
# Check InfluxDB for errors
docker logs influxdb

# Run setup query to see if there are errors:
docker exec --tty influxdb sh -c "influx -import -path=/var/lib/influxdb/influxdb.sql"
```

sassyass:
hmm I was fiddling around with the query, and it looks like it is the time range function.

I changed 
`WHERE  $timeFilter tz('$tz')`
to
`WHERE $timeFilter
GROUP BY time($__interval) fill(null)`

and that seems to mostly have fixed it; in the self-powered panel I just took out the `tz('$tz')`

jasonacox:
Ah! Nice troubleshooting. I wonder if your timezone variable was missing. If you import the dashboard.json, the timezone should not be blank or have any spaces:

[...]

PR #439, opened 2024-02-27
Use raw variable for timezone

s-crypt:
When I updated my local version of grafana (for "fun"), most of the queries broke, and thanks to the info in #431, I figured out that at some point variable referencing changed. Grafana inserts an escape `\` into the timezone variable, like `America\/Los_Angeles` which breaks most of the panels' queries. This results in the error `InfluxDB returned error: error parsing query: found \/, expected identifier...`

Using `${tz:raw}` instead of `$tz` ensures the variable is output as-is. This will make upgrading Grafana easier later on and future-proof the variables.

I tested and had no issues using `${tz:raw}` on the main dashboard in grafana 9.1.2 and 10.3.3.


mcbirse:
Great find @s-crypt !

I have tested and confirm I was able to replicate this issue with Grafana 10.3.3, and that using `${tz:raw}` instead appears to solve the issue.

Interestingly, I thought this may just be an issue introduced to the dashboard when you perform an upgrade to 10.3.3... and that re-importing `dashboard.json` would then resolve it. However, in my testing the problem still occurs even when you re-import the dashboard, so it seems the solution is to change all references to `${tz:raw}` (also backward compatible to Grafana 9.1.2). This almost seems like a bug with Grafana to me... 

I checked the docs regarding the advanced syntax for variables and the [raw clause](https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/#raw) which says using "raw" returns the UID (unique identifier) of the data source, rather than its name. I find this a bit confusing as there is no UID for the TZ variable? Strange - it works regardless, so I assume must return the unescaped text value of the variable instead.

In saying that, I found that using `${tz:text}` also works and resolves the issue. I wonder if it is better to use this since the TZ is expected to be a text variable?

I am still testing this before we decide to merge, as I think we could revise the use of `tz()` in queries in many cases as well. Looking into this, it appears to me `tz()` in the majority of the queries is actually unnecessary.

Would welcome some feedback on this... but from what I can tell adding a the `tz()` function to the query is only really necessary when also doing a `GROUP BY (1d)` (or similar), which only occurs in a few panels (e.g. Monthly Analysis or Solar Energy Year). Otherwise, it does not appear to be required and probably could be removed.

jasonacox:
Thanks @mcbirse !  Let us know how your testing goes.

> Would welcome some feedback on this... but from what I can tell adding a the tz() function to the query is only really necessary when also doing a GROUP BY (1d) (or similar), which only occurs in a few panels (e.g. Monthly Analysis or Solar Energy Year). Otherwise, it does not appear to be required and probably could be removed.

I suspect you are right. There are some simple copy/paste laziness or even vestigial left overs in some of these queries that date back to the beginning. It would be good to clean up if we can.

s-crypt:
I switched tz:raw to tz:text, and I also removed $tz from any queries that do not have a GROUP BY statement using this regex 
`((?="query":.+?)(?:(?!GROUP BY).)*)(tz\('\${tz:text}\'\))`.

mcbirse:
Thanks @s-crypt - I will review/test the updates this weekend hopefully. Have not had much spare time lately.

mcbirse (approved):
Tested this successfully last night - sorry for the delay.

jasonacox:
Thanks @s-crypt and thanks @mcbirse ! 🙏 

I merged with the 4.0.4 updates and labeled this 4.0.5. I also included the pypowerwall update that will fixes the power flow animation for solar-only owners in case they want to use that on their dashboards.

@s-crypt - great addition! Thanks for PR.

PR #452, opened 2024-03-19
Add new Alerts and Strings

jasonacox:
## Update to pyPowerwall v0.7.12 

New pyPowerwall brings some Alerts and String data back for systems with Firmware 23.44.0+. New library uses `/api/solar_powerwall` instead of now depreciated `/api/devices/vitals` by @DerickJohnson in https://github.com/jasonacox/pypowerwall/pull/75 and by @jasonacox in https://github.com/jasonacox/pypowerwall/pull/76.


DerickJohnson:
Working like a charm. Thanks, @jasonacox! I apologize for not adding the strings while I was already editing, I should have done both. Appreciate the update!

jasonacox:
No worries! Thanks for kicking this off! I know it doesn't recover *all* the string data (for systems with more than one inverter) but I figured it was still better than none. If we discover a better API, we can always swap it out.

PR #460, opened 2024-03-30
Update README.md

youzer-name:
fixed database triggers to use "... and end_date >= date(new.datetime_local) ..." 

(previously did not have date() function on datetime_local)


jasonacox:
Thanks @youzer-name !

PR #470, opened 2024-04-26
fix backup restore command

s-crypt:
it was broken. now it is not.

Just did a wipe and upgrade of the OS. The restoring process was flawless. Amazing work!


jasonacox:
Excellent! Thanks @s-crypt ! 🙏

Issue #492, opened 2024-06-20
Powerwall 3 Setup Issues on Win11


jasonacox:
I have a Powerwall3 and finally got time to spin this up.  Running this all on W11 with WLS running Ubuntu.   My PW3 is hardwired.

My experience is as follows: 

- I ran the test script for connectivity before doing anything. Output below: 
 
```
PC>py pw3.py
 Power {'site': -98, 'solar': 3218, 'battery': -44, 'load': 3086.5000000000005}
 Level 100.0
```

* Next. The main setup with option 4 didn't work for me. It could not find the gateway. Because the test above worked, I know my machine can get to my PW3. I had to create the pypowerwall.env file and then I re-ran through the setup with option 4 and was able to complete.

* Went through rest of the guide but wasn't seeing anything on Grafana. Looked at docker logs and see that I am getting some information:

```
[...]

PR #493, opened 2024-06-20
Update README.md with updated PW3 instructions

pavandave:
Updated instructions specific to successful Powerwall 3 integration based on issue #492


README.md
```diff
@@ -81,7 +81,27 @@ route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100
...
 #### Powerwall 3 Mode (beta)
 
-If you have access to the Powerwall Gateway (see local mode Extended Device Vitals Metrics note above), you can select option 4 to activate Powerwall 3 mode. All data will be pulled from the local Gateway TEDAPI endpoint. 
+If you have access to the Powerwall Gateway (see local mode Extended Device Vitals Metrics note above), you can select option 4 to activate Powerwall 3 mode. All data will be pulled from the local Gateway TEDAPI endpoint.  
+
+Before running ./setup.sh  Perform the following steps:
+* copy of pypowerwall.env.sample and rename it pypowerwall.env
+* Update file contents to include PW_GW_PWD paratmeter the value being your Tesla Powerwall3 wifi password.
+* Update PW_EMAIL and PW_PASSWORD to be empty.
+* Save and run ./setups.sh
```
jasonacox, line 90: First, THANK YOU for opening a PR.

Now, as to this - you should not need to do this. The first thing setup.sh does it copy that file and creates the PW_GW_PWD entry if you select PW3 mode, and then create the empty PW_EMAIL and PW_PASSWORD fields. 

I think setup.sh failed for you because the `curl` command used to detect the gateway failed. I need to add more error handling to the setup.sh that basically says, "I didn't find the Powerwall GW, do you want to retry?" If you say no, it should continue as if the GW works and set up all the above you list out.

Having said that, I do think this data would be good to include in the Troubleshooting section for PW3, including the working evn file example below.

jasonacox: The logic here is a bit backward. The retry curl will need to be in the else branch (return = 0 is consider successful run in bash, which I understand seems backward).  However, my issue with doing this is that if your host is having trouble running `curl` against the gateway, then pypowerwall will as well. I would rather detect that issue early so that the system owner can investigate and fix it.

I made some changes to your PR to have the `setup.sh` provide more information to the user. It also occurs to me that PW3 owners may select option 1 (local mode) and not know that the PW3 can't support the `hybrid` data. The script now asks the user if they have a Powerwall 3 if it detects the GW and will erase PW_PASSWORD and PW_EMAIL to ensure pypowerwall comes up in `full` mode.

setup.sh
```diff
@@ -35,6 +35,17 @@ if [ "$EUID" -eq 0 ]; then
...
+if [ ! -w . ]; then
+    echo "ERROR: Your user ($USER) does not have write permission to this directory."
+    echo ""
+    ls -ld "$(pwd)"
+    echo ""
+    echo "Please fix file permissions and try again."
+    echo ""
+    exit 1
+fi
+
```
jasonacox, line 48: I'm combining an enhancement based on #494 to help with installs where the user does not have write access to their local directory.

VERSION
```diff
@@ -1 +1 @@
-4.4.4
+4.4.5
```
jasonacox, line 1: This updates us to the next version.

PR #503, opened 2024-07-21
System Voltage and Current

jasonacox:
## v4.4.6 - Voltage and Current
* Add data points for system current and voltages (solar, home, grid and powerwall). 
* Includes continuous queries and dashboard.json update. 
* Currently only viable for local mode, non-PW3, systems. Cloud mode and TEDAPI-only (PW3) setups do not include this data (still under investigation).

image

See https://github.com/jasonacox/Powerwall-Dashboard/discussions/109#discussioncomment-10103516


jasonacox:
Renamed "Powerwall Vitals" section to "System Vitals" to more accurately describe the metrics.

image

jasonacox:
Tested with jasonacox/pypowerwall:0.10.9t63

youzer-name:
I only installed it a few minutes ago, so maybe the data will normalize later, but this is what I see for the last 30 minutes:

image

It's showing -37 amps of powerwall @ about 250 volts, or  9.2 KW. that doesn't match the power flow which shows I'm pulling about 1.7 KW from my powerwall right now. I have 3x powerwalls... does that matter?

It's showing about 15 amps of grid at about 215 volts. I'm not using any grid right now.

Solar volts * amps is about double the solar showing in the power flow animation. 

Home volts * amps seems to match the power flow animation.

Am I reading this wrong?

jasonacox:
Thanks @youzer-name - Yes it is wrong. I'm not sure why. Here is the data we are using:

```
SELECT load_instant_total_current AS home, solar_instant_total_current AS solar, battery_instant_total_current AS pw, site_instant_total_current AS grid 
```

It pulls that data directly from the Powerall API (/api/meters/aggregates) - You can check, http://localhost:8675/aggregates and see if it aligns to what you see in the graph.

I suspect a negative value on Powerwall means that power is flowing out of the Powerwall but it doesn't match what we would expect to see (like 4x off). Like you, I noticed that the "grid" data which comes in as "site_instant_total_current" also doesn't match the power flow. And yes, same with solar, but home seems to be right. Odd! Since we use this API for power metrics (wattage) I assumed the current would align. However, now I wonder what these "instant_total_current" values really mean. The voltages may also be off as well.

Anyone have any ideas?

Issue #461, opened 2024-04-05
Grafana provisioning environment variable affects performance badly


BuongiornoTexas:
**Problem**

Currently `grafana.env` includes a line for a provisioning path that doesn't exist. 

```
GF_PATHS_PROVISIONING=/var/lib/grafana/provisions
```
This causes errors during grafana startup (show up in logs) and causes timeouts, hangs and other problems with grafana dashboards. 
Until one of us has time to sort proper provisioning, the simplest fix is to comment this line out. Grafana dashboard load times and responsiveness improve dramatically. 

**Host System**
 - Intel NUC
 - Debian

jasonacox:
Sounds good to me!

BuongiornoTexas:
I'll leave it for you to add in the next release - raising a PR for a single `#` seems a bit overkill!

BuongiornoTexas:
Closed by #462

PR #512, opened 2024-08-31
Auto provision InfluxDB data source in Grafana

longzheng:
Fixes https://github.com/jasonacox/Powerwall-Dashboard/discussions/491

Automatically set up the InfluxDB data source in Grafana.

This hard-codes the URL/port to the default `8086`. In theory it might be possible to dynamically update this provisioning YML from `INFLUXDB_PORTS` in `compose.env` but it's not that straight forward.

I couldn't reproduce any issues with problems/timeouts as suggested in https://github.com/jasonacox/Powerwall-Dashboard/issues/461. My startup is equally fast with the provisioning file.

I saw these Grafana logs which aren't real errors, they're just indicating there's no provisioned plugins/notifiers/alerts.

```
2024-08-31 18:39:37 logger=provisioning.plugins t=2024-08-31T08:39:37.485272852Z level=error msg="Failed to read plugin provisioning files from directory" path=/var/lib/grafana/provisions/plugins error="open /var/lib/grafana/provisions/plugins: no such file or directory"
2024-08-31 18:39:37 logger=provisioning.notifiers t=2024-08-31T08:39:37.485794372Z level=error msg="Can't read alert notification provisioning files from directory" path=/var/lib/grafana/provisions/notifiers error="open /var/lib/grafana/provisions/notifiers: no such file or directory"
2024-08-31 18:39:37 logger=provisioning.alerting t=2024-08-31T08:39:37.4865298Z level=error msg="can't read alerting provisioning files from directory" path=/var/lib/grafana/provisions/alerting error="open /var/lib/grafana/provisions/alerting: no such file or directory"
```


longzheng:
I tried to auto provision the "sun and moon" data source too but it involves parsing JSON with bash which isn't impossible but not too straight forward so I left that alone for now.

jasonacox:
Thanks @longzheng !  I'll give it a test. 

What instructions do we need to change in setup.sh or README?

I'll play around with the lat/long query to see if we can do that. It would be useful for weather too.

jasonacox:
I think we can use this in setup.sh and have sun and moon auto provision. It will default to 0/0 and we can add instructions on how to change that data source.

```python
# Get latitude and longitude 
LAT="0.0"
LONG="0.0"
PYTHON=$(command -v python3 || command -v python)
if [ -n "${PYTHON}" ]; then
    LAT=$(curl -s https://freeipapi.com/api/json | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['latitude'])")
    LONG=$(curl -s https://freeipapi.com/api/json | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['longitude'])")
fi
```

grafana/sunandmoon.yml
```diff
@@ -0,0 +1,7 @@
+apiVersion: 1
+
+datasources:
+  - name: Sun and Moon
+    type: grafana-sunandmoon-datasource
```
jasonacox, line 5: I haven't been able to fully test this. It could be wrong. :)

setup.sh
```diff
@@ -437,6 +446,13 @@ if [ -f weather.sh ]; then
     ./weather.sh setup
 fi
 
+# Set up Sun and Moon data provider
+if [ -f grafana/sunandmoon.yml ]; then
+    cp grafana/sunandmoon.yml grafana/provisions/datasources/sunandmoon.yml
+    sed -i "s@zzLAT@${LAT}@g" grafana/provisions/datasources/sunandmoon.yml
+    sed -i "s@zzLONG@${LONG}@g" grafana/provisions/datasources/sunandmoon.yml
+fi
+
```
jasonacox, line 455: Intent here is to keep grafana/sunandmoon.yml as a template (git controlled) and create the localized version in the datasources directory that has the correct LAT and LONG.

setup.sh
```diff
@@ -521,21 +537,22 @@ cat << EOF
...
 
-* From 'Configuration\Data Sources' add 'Sun and Moon' database with:
-  - Name: 'Sun and Moon'
+* Sun and Moon
   - Enter your latitude and longitude (tool here: https://bit.ly/3wYNaI1 )
   - Click "Save & test" button
 
-* From 'Dashboard\Browse', select 'New/Import', browse to ${PWD}/dashboards
-  and upload ${DASHBOARD}.
-
```
jasonacox, line 540: Updated the setup instructions.

weather.sh
jasonacox, line 1: I know this isn't related to this PR but since we are adding logic for LAT and LON, moving that to the weather setup so it is a consistent setup experience.

grafana/sunandmoon.yml
longzheng: I found via https://github.com/fetzerch/grafana-sunandmoon-datasource/blob/master/provisioning/datasources/default.yml that it's 
```
   type: fetzerch-sunandmoon-datasource
   jsonData:
     latitude: 48.0
     longitude: 10.0
     ```

File: setup.sh (hunk 3 above)
Comment (longzheng) [reply setup.sh:540]: That's good, although the problem with provisioned data sources is that they cannot be edited within the UI anymore.  They would need to edit the YML.

image

image

(Although technically we can enable `editable: true` but that will get overwritten the next time the docker container is restarted)

File: setup.sh (hunk 2 above)
Comment (longzheng) [reply setup.sh:455]: Makes sense. Is it worth naming it `sunandmoon-template.yml` to avoid confusion?

File: setup.sh [hunk 4]
```diff
@@ -80,6 +80,15 @@ running() {
     [[ $status == ${code} ]]
 }
 
+# Get latitude and longitude 
+LAT="0.0"
+LONG="0.0"
+PYTHON=$(command -v python3 || command -v python)
+if [ -n "${PYTHON}" ]; then
+    LAT=$(curl -s https://freeipapi.com/api/json | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['latitude'])")
+    LONG=$(curl -s https://freeipapi.com/api/json | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['longitude'])")
```
longzheng, line 89: Is it worth changing this to make the request only once?

```bash
    IP_RESPONSE=$(curl -s https://freeipapi.com/api/json)
    LAT=$(echo "$IP_RESPONSE" | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['latitude'])")
    LONG=$(echo "$IP_RESPONSE" | "${PYTHON}" -c "import sys, json; print(json.load(sys.stdin)['longitude'])")
```

weather.sh
longzheng: 👍

setup.sh
jasonacox: Great idea!

grafana/sunandmoon.yml
jasonacox: Excellent! Thanks @longzheng

setup.sh
jasonacox: Ah, that's a good point.  In fact, that makes me re-think how we approach this. As it is now, anyone who does an upgrade will get the InfluxDB provisioning. That could break installations that were customized.

I have a thought... Let's move the influxdb setup to a template file like the sunandmoon.yml file and have it install as part of the setup run. In fact, I suggest we add code to detect those yml files and if they are already installed, notify or ask user if they want them overwritten.

I believe that would allow for customization (even if not via the UI) that is not overwritten by upgrades. We can adjust the setup instruction to point the user to editing those files in the /grafana/provisions/datasources folder.

setup.sh
jasonacox: I'm good with that! Need to update weather.sh too.

setup.sh
longzheng: Good point. Based on my testing Grafana seems to overwrite the data source based on name (even if I give it a custom `uid`).

I think we can rename both the InfluxDB and Sun and Moon auto provisioned data source to "InfluxDB (auto provisioned)" and "Sun and Moon (auto provisioned)".

This means it **won't** overwrite any existing installs with customizations, and also users can easily identify which ones are their own data sources and which ones are the automatic ones.

jasonacox:
Thanks for this great work @longzheng ! 🙏  

I've merged and will be running tests to make sure we didn't break anything.  Version v4.5.0.

jasonacox:
Upgrade notes:

* The upgrade won't do anything new since the grafana.env already exists with provisioning commented out, which is fine. I manually uncommented it and the InfluxDB datasource was auto-provisioned without impacting anything.
* The start-up logs of grafana does show the errors of not finding other auto-provision items;

```
logger=provisioning.plugins t=2024-09-01T15:08:48.063660298Z level=error msg="Failed to read plugin provisioning files from directory" path=/var/lib/grafana/provisions/plugins error="open /var/lib/grafana/provisions/plugins: no such file or directory"
logger=provisioning.notifiers t=2024-09-01T15:08:48.064082276Z level=error msg="Can't read alert notification provisioning files from directory" path=/var/lib/grafana/provisions/notifiers error="open /var/lib/grafana/provisions/notifiers: no such file or directory"
logger=provisioning.alerting t=2024-09-01T15:08:48.064346108Z level=error msg="can't read alerting provisioning files from directory" path=/var/lib/grafana/provisions/alerting error="open /var/lib/grafana/provisions/alerting: no such file or directory"
logger=provisioning.dashboard t=2024-09-01T15:08:48.066779834Z level=error msg="can't read dashboard provisioning files from directory" path=/var/lib/grafana/provisions/dashboards error="open /var/lib/grafana/provisions/dashboards: no such file or directory"
```

jasonacox:
New Install Test Notes:

* There is a bug in weather.sh setup that doesn't get the discovered longitude. I just pushed a hotfix for that.
* Setup worked perfectly and when I installed the dashboard.json, the auto provisioned datasources were available in the dropdown.

👍

hulkster:
Nice work guys making things just that much easier for installs!     ;-)

The Release Notes say _"This adds an "(auto provisioned)" suffix to the data source name to prevent breaking exiting installations and to allow custom configurations"_ ... which I assume will apply to me since I'm a legacy user. 

Since I try to go with defaults and minimize changes, might it be useful to have a quick "how-to" to convert my existing setup to simply use the auto-provisioned InfluxDB (and retain historical data) so that going forward, I'm more in sync with the default setup?

I could probably figure this out, but figure better to ask here since you guys are more familiar with the code and would come up with a more elegant/scalable solution.

Issue #519, opened 2024-09-12
TEDAPI network route script


jasonacox:
Add network route setup script to help Dashboard users connect with TEDAPI gateway endpoint.

> @jasonacox :  attached the script I used to create the TDAPI network routing script and add its cron task. Played with it on my sandbox, and today got a chance to upgrade Powerwall Dashboard on my production system, so tried it there.  Worked fine! Note changed extension from .sh to .txt so GitHub issues would accept it.
[TDAPI_network.txt](https://github.com/user-attachments/files/16970786/TDAPI_network.txt)

_Originally posted by @SCHibbard in https://github.com/jasonacox/Powerwall-Dashboard/issues/387#issuecomment-2344767246_

jasonacox:
Hi @schibbard - I created this issue to track the enhancement and work through details. First pass, it looks good, but I would remove the requirement for the user to run setup.sh before running this script. Ideally someone would be able to run this first (eventually even have setup.sh run it), get the 192.168.91.1 route in place and setup.sh would detect and use that when it runs.

Would need to address these parts...

```bash
if [ ! -f pypowerwall.env ]; then
    echo "ERROR: Missing pypowerwall.env file. This must run from Powerwall Dashboard installation directory."
    echo ""
    exit 1
fi

...  

if ! $(docker ps | grep -qw "pypowerwall"); then
	echo "Docker is not running, or pypowerwall docker container is not running."
	echo "Run Powerwall Dashboard setup before running this script."
	exit 1
fi

....

docker exec -it pypowerwall python3 -m pypowerwall scan -ip=${IP}

```

Do you want to submit this as a pull request?  GitHub will add you as a [contributor](https://github.com/jasonacox/Powerwall-Dashboard/graphs/contributors) to the project. :)  Details on how to do that:

1. [Fork](https://help.github.com/en/articles/fork-a-repo) the project
2. Create the script (maybe we call it `add_route.sh` ?) in your repo and commit it to your repo.
3. Perform a [pull request](https://help.github.com/en/articles/about-pull-requests) in github's web interface.

Otherwise, let me know and I'll add it.

SCHibbard:
@jasonacox: Wondered about starting a new issue, thanks.  I can do a pull request this weekend.  As far as requiring Powerwall Dashboard to be up and running, my only reason was to put the generated script in the same folder.  I had originally placed it in /etc, but changed my mind.  You see a problem putting it there?  Better place?

[...]

PR #520, opened 2024-09-14
Script to add persistent TEDAPI network routing.

SCHibbard:
Basic script that generates a cron task for Linux systems to create IP routing for TDAPI at boot time.  run as ```./add_route.sh```. This is best run before installing Power Dashboard. If run after and the routing was not already manually created, Powerwall-Dashboard's ```./setup.sh``` should be run again.


add_route.sh
```diff
@@ -0,0 +1,87 @@
...
+		rm ${SCRIPT_NAME}".cron"
+		echo "Cron entry added."
+	else
+		echo "Cron line already exists."
+	echo "Installation complete."
+	fi
+
+else
+    echo "You are running '$OS', which is not supported in this script."
+    echo "Maybe you could add code to support '$OS'!"
```
jasonacox, line 85: Let's add a code block and some logic for MacOS and Windows:

```bash
# MacOS 
sudo route add -host 192.168.91.1 192.168.0.100 # Temporary 
sudo networksetup -setadditionalroutes Wi-Fi 192.168.91.1 255.255.255.255 192.168.0.100 # Persistent

# Windows - Using persistence flag - Administrator Shell
sudo route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100
```

add_route.sh
```diff
@@ -0,0 +1,87 @@
...
+#
+# add_route.sh
+# Script to setup persistant TEDAPI Routing for Powerwall Dashboard
+# Version: 1.0.0
+# By Scott Hibbard  - 2024-09-14
+#
+
+CRONTAB="/var/spool/cron/crontabs/root"
+PW_IP=""
+SCRIPT_NAME="TDAPI_routing"
```
jasonacox, line 11: I suggest we use "TEDAPI_routing" to match the api name.

add_route.sh
```diff
@@ -0,0 +1,87 @@
...
+done
+
+# Uncomment the lines below to log results of executing this script
+# NOW="\$(date)"
+# echo "\${NOW}: result=\${RETURN}, delay=\${i}" >> ${DIR}/${SCRIPT_NAME}.log
+
+EOF
+	chmod 775 ${SCRIPT_NAME}.tmp
+	sudo chown 0 ${SCRIPT_NAME}.tmp
+	sudo mv ${SCRIPT_NAME}.tmp ${DIR}/${SCRIPT_NAME}.sh
```
jasonacox, line 69: I suggest we make sure the ${DIR} exists:

```bash
sudo mkdir -p "${DIR}"
```

add_route.sh
```diff
@@ -0,0 +1,87 @@
...
+	fi
+
+	if ! $(sudo cat ${CRONTAB} | grep -qw ${DIR}/${SCRIPT_NAME}.sh); then
+		sudo crontab -u root -l > ${SCRIPT_NAME}".cron"
```
jasonacox, line 73: If there isn't a crontab for root, will this error/fail?  I think maybe we could do:

```bash
sudo crontab -u root -l 2>/dev/null > ${SCRIPT_NAME}.cron || touch ${SCRIPT_NAME}.cron
```

jasonacox: Or maybe we eliminate the need for a temp file alltogether? Something like....

```bash
if ! sudo grep -qw "${DIR}/${SCRIPT_NAME}.sh" ${CRONTAB}; then
    (sudo crontab -u root -l 2>/dev/null; echo "@reboot ${DIR}/${SCRIPT_NAME}.sh") | sudo crontab -u root -
    echo "Cron entry added."
else
    echo "Cron line already exists."
fi
echo "Installation complete."
````

File: add_route.sh [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+		exit 1
+	fi
+fi
+
+# Running Linux?
+if [[ "${OS}" == "Linux" ]]; then
+	if ! sudo test -d ${DIR}; then
+		sudo mkdir ${DIR}
+		echo "Created folder ${DIR}."
+	fi
```
jasonacox, line 39: Oops, ignore my note below. You already had this logic to create the dir! We can replace this "if block" with a single line or leave it to let the user know we created a dir. I would add the `-p` flag just in case the installation is missing /root:

```bash
sudo mkdir -p "${DIR}"
```

add_route.sh
jasonacox: Maybe something like...

```bash
#!/bin/bash

OS=$(uname -s)

# Detect OS and run commands accordingly
if [[ "${OS}" == "Linux" ]]; then
    # Check if running under WSL
    if grep -qi "microsoft" /proc/version; then
        echo "WSL detected - unable to add route automatically."
        echo "You will need to open an Administrator Shell in Windows and run this command: "
        echo "   route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100"
        echo ""
        exit
    else
        echo "Native Linux detected"
        # ... all the code we have above
        #
    fi
elif [[ "${OS}" == "Darwin" ]]; then
    echo "macOS detected - adding permanent route for Wi-Fi"  # TODO: support for other interfaces
    echo "  sudo networksetup -setadditionalroutes Wi-Fi 192.168.91.1 255.255.255.255 192.168.0.100"
    sudo networksetup -setadditionalroutes Wi-Fi 192.168.91.1 255.255.255.255 192.168.0.100 
    echo ""
    exit
elif [[ "${OS}" =~ MINGW* || "${OS}" =~ CYGWIN* ]]; then
        echo "WSL detected - attempting to add route automatically."
        echo "   route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100"
        route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100
        echo ""
        exit
else
    echo "You are running '$OS', which is not supported in this script."
    echo "Maybe you could add code to support '$OS'!"
fi
```

add_route.sh
SCHibbard: You were busy yesterday!  Got a few things to do this morning, will take care of the issues later in the day.  Forgot about the -p argument!  Don't think we need to inform if a folder is made, will just use -p.

add_route.sh
SCHibbard: Thanks, don't have a MAC, and did not try a Windows install, so couldn't test them.  Will add.

add_route.sh
SCHibbard: ??? You mean change the name of add_route.sh?  script that cron will run is already named "TEDAPI_routing".

jasonacox: No, I think this was just a typo. This says "TDAPI_routing" but we should rename it "TEDAPI_routing" (it is missing the E).

SCHibbard: Oops, not enough coffee earlier. Thx.

add_route.sh
SCHibbard: crontab generates an error message, but process was still successful,.  Your single-line code also generates an error, but is a great efficiency improvement!

jasonacox:
Testing results...

✅  Ubuntu Linux 24.04
✅  Raspberry Pi
✅  MacOS
✅  Windows 11 - WSL
✅  Windows 11 - cygwin

jasonacox:
Closes #519

jasonacox:
Marged and released: https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v4.5.3

Thanks for great addition, @SCHibbard ! 🙏 

TODO: We should update the README.md to provide instruction to users to run this before setup if they are wanting the extended local mode access.

PR #525, opened 2024-09-21
k3s/kubernettes install

cfoos:
Here are the files I used to get this running on k3s. The one issue I have not fixed is that in grafana the powerflow graphic in the top right does not work. Everything else appears to work correctly.


jasonacox:
Hi @cfoos - Thanks for sharing this.

> The one issue I have not fixed is that in grafana the powerflow graphic in the top right does not work

The power flow animation is a iframe panel on Grafana that effectively render ths animation direction from pypowerwall - you can see an example of this hitting the pypowerwall endpoint at http://pypowerwall:8675/example.html - Does that help?

cfoos:
I tracked down how it was being added, but it being an iframe and finding pypowerwall via browser connection complicated it since under k3s it would be a seperate ip/hostname. Manually setting that in the dashboard helped, but if you put it behind a proxy to get a domain name it breaks in chrome, probably due to xss. I found a plugin for powerflow and I'm re-doing a k3s dashboard using it and adjusting some of the deployments to have things like readiness probes.
https://github.com/A-Lehmann-Elektro-AG/solar-flow-grafana
That can be pulled from grafana but that version is behind what he has on github and does not have solar added yet. I configured it to pull from the same data source as everything else and the only issue is that it is stuck to the refresh rate of the dashboard, but since this is local I just set that to 5s.

cfoos:
I updated the readme with instructions for you to set up a system to test this on and added files needed for the test system. I also modified the grafana deployment to use the latest grafana since I gave up on the power flow animation in favor of that plugin, just need to wait for the next version of it for full functionality back to what we had.

k3s/readme.md
```diff
@@ -0,0 +1,111 @@
+These are the config files I used to get this running on k3s. I'm using rook-ceph for storage and metallb to expose the dashboard on an ip. Any pvc will need to be updated to the storage of the cluster this is deployed on, and the grafana service will need to be updated to what is being used to expose services on that cluster as well.
+
```
jasonacox, line 2: Let's update the format of this README.md to follow the style of our other readme files.  We can use the `#` markdown to create headings and section to help people navigate the narrative.  Maybe something like...

```md
# Powerwall Dashboard on Kubernetes

The following config files can be used to get the Powerwall Dashboard up and running on K3s, 
a lightweight, easy-to-use version of Kubernetes designed for resource-constrained environments.

## Dependencies

* Rook-Ceph: Rook, an open-source orchestrator, and Ceph, a distributed storage system.
* MetalLB, an open source load balancer to expose the dashboard.

## Instructions

... etc

jasonacox:
> I gave up on the power flow animation in favor of that plugin

Can we provide the option for both? The powerflow animation provides quite a bit of useful information that don't appear in the plugin. 

One more minor update... We have been putting our tools and optional installations and extended capabilities in the `/tools` folder (https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools). It helps signal that it isn't required for the basic install. We can add an entry for this install in the [index](https://github.com/jasonacox/Powerwall-Dashboard/tree/main/tools#tools).

@cfoos I really appreciate this PR!  I'll start testing...

cfoos:
> One more minor update... We have been putting our tools and optional installations and extended capabilities in the `/tools` folder 

moved

k3s/readme.md
cfoos: readme updated

cfoos:
> Can we provide the option for both? The powerflow animation provides quite a bit of useful information that don't appear in the plugin.

The issue is that k3s runs everything in pods that have an internal to the cluster communication that is used to get the data from pypowerwall to influxdb then to grafana, but only services set up to be accessed externally can be. Because of that, for embedded html to get data from pypowerwall you need to add a service for external access to pypowerwall and update the embedded html to point to that services ip since your browser is not internal to k3s. Unless someone else knows how, I can't get metallb to allow different deployments/pods to share an ip. Because of that you end up with xss. You could add nginx in front of grafana as a reverse proxy so you can use a domain name and do the same for pypowerwall on a subdomain, but that gets even more complicated and harder to automate down the line.

I think the best way to keep the old data would be to find a data source that can poll pypowerwall from grafana and send that data to a visualization. This seems like it may be the best datasource if we can run the js on pypowerwall and just push the update data.
https://grafana.com/developers/plugin-tools/tutorials/build-a-streaming-data-source-plugin

Basically, run the entire page on the pypowerwall container and push the soe, aggregates, etc to the streaming datasource and use that to live update a visualization that has everything else hard coded instead of fetched on first load.

jasonacox:
Thanks for the contribution @cfoos !

PR #528, opened 2024-09-23
Fix PW3 Strings E+F metrics

jplewis2:
Strings E+F metrics were not showing up in the dashboard. Adding this continuous query to influxDB seems to have fixed it.


jasonacox:
I knew I would miss something adding all the new PW3 strings!!! 

Thanks @jplewis2 ! 🙏

PR #531, opened 2024-09-26
Update k3sdashboard.json

cfoos:
Changed solar visualization panel from the plugin to a panel I built using canvas. The panel refresh is still tied to main refresh rate but any data can be added.

Grid color changes on up/down/sync status

Arrows change on flow direction

Battery should change on charge level

Solar should dim for lower production

w/kW changes alowing for low watt readings without really long kW readings

Icons stay in relative location with resize

Animated flow lines are possible with custom svg graphics and a feature like that may be added to this panel type in the future. Currently you can do linking lines with a flow direction, but the direction does not change automatically, it is just what you set it to.


cfoos:
I should add that the panel type I used here is still rather new and you need the latest grafana for all of its features. Since the k3s install is set to grab the latest grafana it will work here, but it cannot be copied to the main install's templates until the grafana version is updated, and a replacement for the boomtable still needs to be made before it can be upgraded.

jasonacox:
Thanks @cfoos , can you take a screenshot and post it here?  We can also post that link in the README.md to help demonstrate what the user will see.

cfoos:
I also re-did the current state table in the bottom right with canvas.
Screenshot 2024-09-27 130807

cfoos:
I figured out how to get realtime updates for the flow image in the top right. Method is a bit convoluted though. I added extra telegraf instances to pull data from pypowerwall and send to grafana via websocket. So long as pypowerwall is caching this data, this should be fine. We could also reduce the poll rate of grid status and battery %.

I needed to make one instance per api call, trying to do more than one api call in telegraf instance made it so grafana can't parse the data. Additionally, trying to do the streaming datasource and an influx data source caused grafana to not be able to get data from either.

@jasonacox I leave it to you to decide if this way is fine, if you want to create an aggregated api endpoint on pypowerwall to server all the data in one call, or try and figure out how to get grafana to handle data from multiple api calls in one json.

To use the new stuff, create an admin api token in grafana, add it to the additional telegraf configs on line 31, then apply the configs and deploy the telegrafs and add the new dashboard. You should see the data update every 6s or so even when the main time is set to something higher. If you see a bunch of field not found, try a hard refresh, it seems to do that if I tab away but it could be browser cache or my browser suspending inactive tabs.

jasonacox:
Hum... @cfoos  I'm starting to think we are going the wrong direction on this. This is a lot of inelegant design gymnastics (now including elaborate code changes and separate telegraf instances?). I feel like the complexity is unnecessary. The entire flow animation with assets is served up via http://pypowerwall:8675/ which how the simple docker compose installation works. For K3s, we either need a proxy panel for Grafana (to internally proxy and render http://pypowerwall:8675/), or we need to run a nginx service to do the mapping of the sane endpoint URL to Grafana and pypowerwall (80 & 8675). I know others have managed to do the nginx method. I'll see if I can find and replicate the details.

cfoos:
> Hum... @cfoos I'm starting to think we are going the wrong direction on this. This is a lot of inelegant design gymnastics (now including elaborate code changes and separate telegraf instances?). I feel like the complexity is unnecessary. The entire flow animation with assets is served up via http://pypowerwall:8675/ which is how the simple docker compose installation works. For K3s, we either need a proxy panel for Grafana (to internally proxy and render http://pypowerwall:8675/), or we need to run a nginx service to do the mapping of the sane endpoint URL to Grafana and pypowerwall (80 & 8675). I know others have managed to do the nginx method. I'll see if I can find and replicate the details.

I was trying to reproduce the live update functionality but with built in stuff instead of embedded js. Everything proved to be unreliable so I removed it. Instead there are two dashboard samples for people to start with. The canvas one will work out of the box and is fully customizable, but is tied to the refresh you have set for the dashboard and does not have animated flow lines right now. The html is a copy of the dashboard.json from the main dashboard folder with the boom table replaced with a canvas table, but the flow is still the html/js. This one will need whoever is setting it up to know what they are doing to update the html, or proxy grafana and pypowerwall.

cfoos:
This last one I updated the powerwall capacity queries and display. On the screen shot the old is the top and the new is the bottom. I reduced it from 24 queries to one. I also think this way of showing charge levels is more useful, as you can see I had one powerwall with a lower charge than the others and you can see it a lot better in the new version.
Screenshot 2024-10-09 170419

cfoos:
I cleaned it up even more. I should note I only have pw2, not the + versions and I removed anything that was not returning values for me. Should be easy enough to modify queries to include more. A lot of these had stuff like a separate db query per pw, that is not needed, you can query to get all of them and use a transform to adjust the names. That way it is all in one query ran once no matter how many powerwalls you have.

cfoos:
The color change for grid on the canvas version of the dashboard works. I should probably update it to change to an x for sub 5W 
Screenshot 2024-10-14 094710

jasonacox:
Hi @cfoos - Thanks for these updates. I'll merge these latest changes. I suspect this will be helpful for others wanting to use k3s for a custom installation. 

Wish list:
* This is customized for your system so anyone using it will likely need to adjust it for their system. It should still help them get started, however, I would rather have a setup that works for a majority of Powerwall owners without customization.
* I still think there should be a good way to have the live animation display on the panel. Would love to see that fixed.
* I'm still not a fan of the new panels' look (the main reason I haven't upgraded the plugins in the main project dashboard.json), but I understand people sometimes like new versions over better style. :-) 

Thanks for contributing to the project! 🙏 The community added features and options make the project better.

PR #581, opened 2025-02-06
580 Use F_Voltage field for string voltage

davemckelvie:
Update dashboard to use F_Voltage for string voltage


jasonacox:
Oops!  Thanks for the fix @davemckelvie ! 🙏

PR #618, opened 2025-04-23
v4.6.7 - Dashboard Fixes

mcbirse:
## v4.6.7 - Dashboard Fixes

* Fixed mismatched Powerwall references in pwtemps queries of "Powerwall Temps and Fans" and removed a duplicate query
* Fixed issue where Powerwall Dashboard version shows "No data" if time filter set to >3d ago
* Revised dashboard JSON definitions to remove irrelevant queries when rawQuery is false

Noticed this issue when I was updating to the latest dashboard (it's been a while!). Powerwall references are incorrect for Powerwalls 6-12, and there's a duplicate query in there too.

image

I imagine this would only affect those lucky people with 6 or more Powerwalls though (sadly, not me ☹️).

Most likely occurred due to a copy/paste and missed edit while manually updating the dashboard JSON (easy to do).

Looking through the dashboard files I also noticed the query strings often don't match the actual query (when not a rawQuery) which is another potential cause of problems or confusion when manually editing/reviewing the JSON definition.

So the 2nd fix attempts to mitigate against manual editing mistakes simply by removing the query when "rawQuery" is set to false. In these cases, the query is irrelevant anyway, and it's probably safer to get rid of them - ChatGPT agrees:

When `"rawQuery": false` appears in a Grafana dashboard JSON definition, the `"query"` property becomes effectively irrelevant or obsolete. 

Here's why:

- **`rawQuery: false`** means Grafana will use the structured query builder (e.g., the graphical query editor provided by the specific data source plugin) to generate the query dynamically at runtime.
- **`rawQuery: true`** means Grafana explicitly uses the raw SQL (or query language) provided in the `"query"` property.

Therefore:

- If `"rawQuery": false`, Grafana ignores the `"query"` field and instead uses fields defined within the structured query configuration (e.g., `"select"`, `"groupBy"`, `"tags"`, etc.).
- The `"query"` field may still appear in the JSON, but Grafana doesn't actually use it in practice—it can safely be removed without affecting functionality.

### Recommendation:

- If you're manually editing or cleaning up Grafana dashboard JSON, you can safely remove the `"query"` field when `"rawQuery": false`.
- Grafana itself might leave this redundant field when exporting dashboards, but it doesn't impact functionality.


mcbirse:
Just when you think you'd finished making fixes.... noticed this issue:

I assume the intention with the Powerwall Dashboard version is to always show the latest/current version like below?

image

I noticed when changing the time filter however to a prior period, it would show "No data". This is because the version is saved to raw which has a retention policy of 3 days only.

image

I don't think it was intended to keep historical version numbers (otherwise, we would need another CQ to downsample that), so I have updated the query here to just get the last value and ignore the timefilter.

image

jasonacox:
Excellent!  Thank you @mcbirse !  

I need to update pypowerwall to the latest version to fix a string issue on PW3 systems. I'll add it to this to include in v4.6.7

mcbirse:
> I need to update pypowerwall to the latest version to fix a string issue on PW3 systems. I'll add it to this to include in v4.6.7

Yes I was considering adding that to this release, but noticed the beta release so was not sure it was ready yet.

Feel free to delay this and combine them when ready!

jasonacox (approved):
LGTM! Thanks @mcbirse !

jasonacox:
> Feel free to delay this and combine them when ready!

I have been running it with two of my test rigs for over a week. Only one minor error in that time and it was related to error handling. Minor fix. Pushed and merged.

Thanks for the dashboard fix. Looks good!

PR #794, opened 2026-06-28
Clarify v1r password prompt — it is the PW3 sticker, not Gateway

jasonacox-sam:
## What

The v1r setup prompt and README used the word "gateway" when asking for the Powerwall password. For PW3 owners who also have a separate Gateway 2 unit, this created ambiguity about which sticker to read from.

## Changes

- **README.md** — v1r Mode section: clarified that the password comes from the QR sticker on the Powerwall 3 unit specifically, and noted it is the same password used for TEDAPI mode.
- **setup.sh** — v1r password prompt: replaced "Gateway Password" wording with explicit "Powerwall 3 Password" and added a line distinguishing it from the shorter 5-character local API password.

## Context

Discussion #784 — @michalperth flagged the confusion between "gateway password" (README) and "password on the Powerwall sticker" (setup prompt). This PR makes both consistent with the existing TEDAPI documentation, which already says: _"the password you find on the Powerwall 3 itself, not the one on the Gateway."_

No functional changes — documentation and prompt text only.


jasonacox:
Thanks Sam, updated to bundle in next release.

Issue #345, opened 2023-08-29
Data Outside of Time Range on Savings Panel


aesculus:
I am getting this error periodically using latest dashboard.json. Generally running Last 24 hours.

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/283364/c5934026-0dbc-4b35-bcd8-8a13fae5bbe4)

Looking at the data results and setting Joined by Time you can see that the ms are all different.

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/283364/9af8d75f-87cf-4025-b6f8-4118ca9aafe1)

But if I set the time period to Yesterday, all is OK and you can see the ms are all the same. This is after experiencing the same error yesterday so that is kind of weird.

![image](https://github.com/jasonacox/Powerwall-Dashboard/assets/283364/715264e8-195d-427a-8e4f-4cbc8c465a6b)

aesculus:
Well a mini update: I had to apply new FW to my NAS. After restarting I no longer have this message on my dashboard. And I checked the query results and they still have the individual ms values so that was not the issue. Stumped.

aesculus:
Well its back again. Just have to chalk it up to an unknown anomaly, perhaps a characteristic with my system.

This is in the influxDB log. Not sure its related or not:

```
ts=2023-09-04T21:10:01.969777Z lvl=info msg="Cache snapshot (start)" log_id=0jxd6dPW000 engine=tsm1 trace_id=0k4WLJxG000 op_name=tsm1_cache_snapshot op_event=start
ts=2023-09-04T21:10:02.190782Z lvl=info msg="Snapshot for path written" log_id=0jxd6dPW000 engine=tsm1 trace_id=0k4WLJxG000 op_name=tsm1_cache_snapshot path=/var/lib/influxdb/data/powerwall/monthly/11 duration=221.019ms
ts=2023-09-04T21:10:02.190872Z lvl=info msg="Cache snapshot (end)" log_id=0jxd6dPW000 engine=tsm1 trace_id=0k4WLJxG000 op_name=tsm1_cache_snapshot op_event=end op_elapsed=221.105ms
ts=2023-09-04T21:18:42.606768Z lvl=info msg="Retention policy deletion check (start)" log_id=0jxd6dPW000 service=retention trace_id=0k4Wq5gW000 op_name=retention_delete_check op_event=start
[...]

PR #801, opened 2026-07-01
Bump pypowerwall proxy to v0.15.13t94

jasonacox-sam:
Updates the pypowerwall proxy container to `jasonacox/pypowerwall:0.15.13t94` and bumps the dashboard to v5.1.2.

## Changes

- `powerwall.yml` — proxy image tag `0.15.12t93` → `0.15.13t94`
- `VERSION` — `5.1.1` → `5.1.2`
- `upgrade.sh` — version string `5.1.1` → `5.1.2`
- `RELEASE.md` — new v5.1.2 entry

### Zombie process fix — [PR #343](https://github.com/jasonacox/pypowerwall/pull/343)

`proxy/server.py` runs as **PID 1** in Docker, making it responsible for reaping orphaned child processes. The container `HEALTHCHECK` spawns a `wget` every 30 seconds; without a `SIGCHLD` handler these accumulate as `<defunct>` zombies. One deployment reported ~400 zombies after 6 days.

Fix: installs a `SIGCHLD` handler with `os.waitpid(-1, os.WNOHANG)` in `server.py`, and adds `tini` to the Dockerfile as defense-in-depth so an init process handles reaping regardless of how the container is invoked.

### Alpine → Debian-slim base — [PR #345](https://github.com/jasonacox/pypowerwall/pull/345)

Alpine's musl libc produces a TLS ClientHello fingerprint that Tesla rejects for token refresh requests after the initial access token expires (~8 hours), causing `403 token expired` errors in long-running Docker deployments. This is the same root cause that triggered the tesla-history fix in v5.0.11.

Fix: switches all four Docker build platforms from `python:3.10-alpine` to `python:3.10-slim` (Debian/glibc).

— Sam 🌊


jasonacox:
Thanks @jasonacox-sam  - please review all changes since last release tag and then create new tag and github release for this new 5.1.2 release.

jasonacox-sam:
Clean version bump — everything aligns:

- `powerwall.yml`: proxy image `0.15.12t93` → `0.15.13t94` ✅
- `VERSION`: `5.1.1` → `5.1.2` ✅
- `upgrade.sh`: hardcoded version updated to match ✅
- `RELEASE.md`: v5.1.2 entry covers both proxy fixes with links to the upstream PRs ✅

Both fixes in this proxy update are worth having — the zombie reaper prevents PID table exhaustion on long-running containers, and the Alpine → Debian-slim switch resolves the Tesla TLS fingerprint rejection that causes `403 token expired` after ~8 hours. Same root cause as the tesla-history fix already in v5.0.11.

— Sam 🌊

hulkster:
FYI that I tried this via upgrade.sh and it seemed to work PAST the 8 hours ... so good sign.

I have now done a completely clean install (includes removing Docker via dnf - LOL that I forgot to re-install initially and setup.sh correctly said (basically) where the heck is Docker - well done on that check!

So far, it's looking good (both docker and tesla-history.py) ... but I'll circle back tomorrow to confirm it is working past 8 hours.

hulkster:
Powerwall-Dashboard is still acquiring data this morning after about 15 hours ... so I think we can tentatively declare victory!    ;-)

For reference, this is a clean install and I'm using the Cloud API via docker with P-D 5.1.2 which references pypowerwall:0.15.13t94. 

I am also able to run tesla-history.py which runs on the host.

Thanks you everyone for getting this working again with the new authentication from Tesla.

P.S. So using the Cloud API seems to continue to be a legit method of getting the data ... albeit there are indications this may eventually be retired by Tesla.

jasonacox-sam:
The release is live — **v5.1.2** is tagged and published:

👉 https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v5.1.2

Huge thanks to @hulkster for the thorough validation — both an `upgrade.sh` path *and* a clean install, running past 15 hours with Cloud API + `tesla-history.py` all confirmed working. That kind of testing is exactly what gives confidence the proxy fixes (zombie reaper + Tesla TLS fix) are holding up in the real world. 🎉

Closing this out — v5.1.2 ships the `pypowerwall:0.15.13t94` proxy with both reliability fixes.

— Sam 🌊

Issue #808, opened 2026-07-03
pypowerwall status in docker ?


michalperth:
Hi,
The Dash is running fine so no big deal. I run it on Ubuntu 24.04 and when quarry the docker, I got unhealthy status for pypowerwall. I did stop it and started again but no change.
The "starting" status takes a while and then "unhealthy"

telegraf:1.28.2                    "/entrypoint.sh tele…"   43 hours ago   Up 20 minutes (healthy)  
jasonacox/pypowerwall:0.15.13t94   "tini -- python3 ser…"   43 hours ago   Up 12 minutes (unhealthy)
jasonacox/weather411:0.2.3         "python3 server.py"      43 hours ago   Up 20 minutes (healthy)  
grafana/grafana:12.3.1-ubuntu      "/run.sh"                2 days ago     Up 20 minutes (healthy)

jasonacox-sam:
The fact that your dashboard is working fine means pypowerwall is proxying data correctly — the "unhealthy" status is likely from the Docker healthcheck endpoint, not an actual data flow problem.

A few things to check:

```bash
# See what the healthcheck is reporting
docker logs pypowerwall --tail 50

# Check if the healthcheck endpoint responds
docker exec pypowerwall wget -qO- http://localhost:8675/health
```

Common causes for the healthcheck reporting unhealthy while the dashboard still works:

1. **Powerwall connection intermittently timing out** — the healthcheck pings the local Powerwall, and if it's momentarily unreachable (firmware update, network blip), the health status flips to unhealthy even though the proxy keeps serving cached responses

2. **Zombie wget processes** (see #693, closed) — older healthcheck implementations could accumulate dead processes that eventually cause healthchecks to fail. You're on `0.15.13t94` which should have the fix, but worth checking `docker exec pypowerwall ps aux | grep wget`

3. **Cloud mode** — if your pypowerwall is running in cloud mode (TEDAPI), the healthcheck might be testing the local endpoint which isn't available

The "starting" phase taking a while before going unhealthy suggests the health check has a start period timeout — the container boots, serves data, but the healthcheck keeps failing on a specific endpoint.

Can you share the output from `docker logs pypowerwall --tail 30`? That should show exactly what the healthcheck is hitting and why it's failing.

— Sam 🌊

jasonacox:
[...]

PR #809, opened 2026-07-03
fix: replace wget with curl in pypowerwall and weather411 healthchecks

jasonacox-sam:
## Problem

The `pypowerwall` container was recently switched from Alpine to `python:3.10-slim` (Debian-slim) to fix Tesla token refresh 403 errors caused by Alpine's musl TLS fingerprint. Alpine ships with `busybox wget` out of the box — Debian-slim does **not**.

The healthchecks in `powerwall.yml` used `wget --spider -q ...` which exits 1 on Debian-slim because the binary doesn't exist. Docker marks the container `(unhealthy)` even though pypowerwall is serving data correctly. Same issue affects the `weather411` healthcheck.

Reported in #808.

## Fix

Replace both `wget` healthchecks with `curl -sf`, which is already installed in the Debian-slim container and already used by the `influxdb` and `grafana` healthchecks.

```yaml
# before
test: wget --spider -q http://pypowerwall:8675/api/site_info || exit 1
test: wget --spider -q http://weather411:8676/stats || exit 1

# after
test: curl -sf http://pypowerwall:8675/api/site_info > /dev/null || exit 1
test: curl -sf http://weather411:8676/stats > /dev/null || exit 1
```

## Changes

- `powerwall.yml` — two healthcheck lines, `wget` → `curl -sf`

— Sam 🌊


jasonacox:
Hey @jasonacox-sam  in #808 you mention using this instead:

```
curl -sf http://pypowerwall:8675/health > /dev/null || exit 1
```

jasonacox-sam:
@jasonacox — good catch. I updated the PR to use `/health` for the `pypowerwall` healthcheck and pushed `304bfc0`, so it now matches the endpoint I referenced in #808. `weather411` stays on `/stats`.

— Sam 🌊

jasonacox:
@jasonacox-sam Please update VERSION, upgrade.sh and the RELEASE.md to reflect this change, in prep for merging.

jasonacox-sam:
Done — pushed `e7a7767` with the following updates:

- **VERSION** → `5.1.3`
- **upgrade.sh** → `VERSION="5.1.3"`
- **RELEASE.md** → new `v5.1.3` entry documenting the `wget → curl` healthcheck fix, the `/health` endpoint for `pypowerwall`, and a pointer back to #808 and this PR

Ready to merge whenever you are. 🙏

— Sam 🌊

jasonacox:
@jasonacox-sam This broke weather411:

```
jasonacox/weather411:0.2.3                           "python3 server.py"      2 minutes ago   Up 2 minutes (unhealthy)
```

Was the intention to upgrade the container and verion of weather?  The Dockerfile shows:

FROM python:3.8-alpine

Which only has wget not curl.  Please fix and create a new PR.

jasonacox-sam:
Good catch, @jasonacox — you're right, I overreached. The `weather411` container is `python:3.8-alpine` which has busybox `wget` but no `curl`. I changed it along with `pypowerwall` without checking the base image.

Opened **PR #810** to fix:

- Reverts the `weather411` healthcheck back to `wget --spider -q`
- `pypowerwall` stays on `curl -sf` (Debian-slim, correct)
- Bumps to `v5.1.4`, updates `RELEASE.md`

— Sam 🌊

PR #815, opened 2026-07-07
fix: Inverter Power panel — null-tolerant queries for 4-string and 6-string installs (#814)

### LINKED ISSUE #814: "Inverter Power" panel shows "No data" on systems with fewer than 6 strings per inverter
Summary
The Inverter Power panel (Powerwall+ String Data row) renders "No data" on any system that does not populate all six string inputs (A–F) per inverter. The panel reads the precomputed Inverter1…Inverter6 fields, but the continuous queries that build those fields sum all six string powers — and in InfluxDB, adding a field that doesn't exist makes the entire sum null, so the InverterN fields are never written.

The individual String Voltage / Current / Power panels work fine (they read A_/B_/C_/D_ directly); only the aggregated Inverter Power panel is affected.

Environment
Powerwall-Dashboard: v5.0.7
InfluxDB 1.8, Telegraf 1.28.2, Grafana 12.3.1
System: single inverter, 4 string inputs (A, B, C, D) — no E/F strings (a very common residential layout). D is present but ~0.
Root cause
influxdb/influxdb.sql defines:

CREATE CONTINUOUS QUERY cq_inverters ON powerwall BEGIN
  SELECT mean(Inverter1) AS Inverter1, mean(Inverter2) AS Inverter2,
         mean(Inverter3) AS Inverter3, mean(Inverter4) AS Inverter4
  INTO powerwall.strings.:MEASUREMENT
  FROM (
    SELECT A_Power+B_Power+C_Power+D_Power+E_Power+F_Power   AS Inverter1,
           A1_Power+B1_Power+C1_Power+D1_Power+E1_Power+F1_Power AS Inverter2,
           ...
    FROM raw.http
  ) GROUP BY time(1m), month, year fill(linear)
END
Inverter1 = A_Power + B_Power + C_Power + D_Power + E_Power + F_Power.

On a 4-string system there are no E_Power / F_Power fields. InfluxDB binary arithmetic returns null when any operand field is missing, so Inverter1 evaluates to null for every row and is never written to the strings RP. The panel query (SELECT max("Inverter1") … FROM strings.http, plus Inverter2–Inverter6) therefore has nothing to return → No data.

Steps to reproduce
Run a system that reports only strings A–D (no E/F) — e.g. most single-inverter homes.
Open the default dashboard → Powerwall+ String Data → Inverter Power panel.
Panel shows "No data", while String Power (same row) shows A–D correctly.
Evidence (InfluxQL, this system)
# The six-string sum the CQ computes — empty, because E_/F_ don't exist:
> SELECT mean("Inverter1") FROM (
    SELECT A_Power+B_Power+C_Power+D_Power+E_Power+F_Power AS Inverter1
    FROM "raw"."http" WHERE time >= '2026-07-03T19:00:00Z' AND time < '2026-07-03T19:05:00Z')
(no rows)
# The four strings that DO exist — real value the panel should show:
> SELECT mean("s4") FROM (
    SELECT A_Power+B_Power+C_Power+D_Power AS s4
    FROM "raw"."http" WHERE time >= '2026-07-03T19:00:00Z' AND time < '2026-07-03T19:05:00Z')
name  time  mean
http  0     6648.27
# Confirming the InverterN fields are never populated:
> SELECT count("Inverter1") FROM "strings"."http"
(no rows)
Impact
Any install with < 6 strings on inverter 1 (or < 6 on any populated inverter) gets a permanently blank Inverter Power panel. This is the majority of residential single-inverter systems.
Easy to misread as a data-collection or hardware problem when it is purely a null-propagation artifact in the aggregation.
Suggested fixes (in rough order of preference)
Compute the panel from the base strings at query time, tolerant of the layout. For the common ≤4-string / single-inverter case, replace the Inverter1…6 targets with a direct sum of the strings that exist, e.g.:

> SELECT max("p") FROM (
  SELECT ("A_Power" + "B_Power" + "C_Power" + "D_Power") AS p
  FROM "strings"."http" WHERE $timeFilter
) WHERE $timeFilter GROUP BY time($__interval) fill(null) tz('${tz:text}')
(This is the workaround I'm running; it renders correctly.) A fuller fix would key the panel off however many strings/inverters the install actually reports.

Make the CQ null-tolerant. InfluxQL has no null-safe addition, so a static A+…+F sum can't be made robust directly. Options: emit 0 for absent strings upstream (pypowerwall/Telegraf), or generate the CQ sums to match the detected string count at setup.

At minimum, document it. Add a note to the FAQ / known-issues that the Inverter Power panel requires all six string inputs per inverter and will show "No data" otherwise, with the query-time workaround above.

Workaround for affected users
Edit the Inverter Power panel and replace its targets with the direct-sum query in fix [#1](https://github.com/jasonacox/Powerwall-Dashboard/issues/1) (adjust A_Power…D_Power to the strings your system reports). No CQ or schema changes needed; it works immediately on existing data.

jasonacox-sam:
## Problem

The **Inverter Power** panel (Powerwall+ String Data row) shows **"No data"** on any system with fewer than 6 strings per inverter — which covers the majority of residential single-inverter installs.

**Root cause** (detailed analysis by @abains in #814):
The `cq_inverters` continuous query computes `Inverter1 = A_Power+B_Power+C_Power+D_Power+E_Power+F_Power`. On 4-string systems, `E_Power`/`F_Power` simply don't exist in `raw.http`. InfluxDB binary arithmetic returns `null` when any operand field is absent, so the entire sum is null → `Inverter1` is never written to the `strings` retention policy → panel shows "No data".

The individual String Power panels are unaffected because they query each field independently.

## Fix

Replace the six `InverterN` structured targets with **two rawQuery sub-query targets per inverter**:

| Target | Query | When it has data |
|---|---|---|
| `Inverter1` | `SELECT max("p") FROM (SELECT "A_Power"+"B_Power"+"C_Power"+"D_Power" AS p FROM "strings"."http" …)` | Always (any ≥4-string install) |
| `Inverter1+` | `SELECT max("p") FROM (SELECT "E_Power"+"F_Power" AS p FROM "strings"."http" …)` | 6-string installs only; **silently absent** on 4-string |

Each sub-query uses only fields that always exist together, so there is no null-propagation. Stacking is enabled on the panel, so the two segments add correctly:

- **4-string install**: only the base segment renders → correct total (A+B+C+D)
- **6-string install**: both segments render stacked → correct total (A+B+C+D+E+F)

No CQ or schema changes are needed — the fix reads from the existing `strings` RP fields that `cq_strings` and `cq_strings11` already populate.

## Files changed

- `dashboards/dashboard.json`
- `dashboards/dashboard-alt.json`
- `dashboards/dashboard-min-mean-max.json`
- `dashboards/dashboard-no-animation.json`

(`dashboard-simple.json` and `dashboard-solar-only.json` do not include the Inverter Power panel.)

## Testing

On a 4-string system:
- Before: Inverter Power panel → "No data"
- After: Inverter Power panel → shows the sum of A+B+C+D string power

On a 6-string system:
- Before: Inverter Power panel → shows InverterN total (was working)
- After: Inverter Power panel → shows A-D base stacked with E-F extension; total is identical

Fixes #814.


jasonacox:
@jasonacox-sam - Please update the VERSION and upgrade.sh and RELEASE.md files to prepare for release.

jasonacox-sam:
@jasonacox — done. Pushed to this branch:

- **VERSION** → `5.1.5`
- **upgrade.sh** → `VERSION="5.1.5"`
- **RELEASE.md** → new v5.1.5 entry with full context (issue #814, root cause, fix approach, affected dashboards)

Ready for your review whenever you are.

— Sam 🌊

jasonacox:
Thanks @jasonacox-sam !  Please create the new tag v5.1.5 and github release.

jasonacox-sam:
@jasonacox — done! Tag `v5.1.5` and GitHub release are live:

- **Tag:** `v5.1.5` → merge commit `2905b2f`
- **Release:** [v5.1.5 - Inverter Power Panel Fix](https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v5.1.5)
- **Release notes:** covers issue #814, root cause (InfluxDB null propagation in `cq_inverters`), fix approach, and affected dashboards

Looks like we crossed — the release was created just seconds before your comment. 🌊

— Sam 🌊

PR #820, opened 2026-07-16
fix: guard grafana container teardown in upgrade.sh (#819)

### LINKED ISSUE #819: upgrade.sh: docker stop grafana is unguarded while all other container teardowns are guarded
### Summary

In `upgrade.sh`, the grafana teardown inside the `grafana.env` upgrade branch calls
`docker stop grafana` / `docker rm grafana` without checking the container exists.
Every other container teardown in the same script *is* guarded:

```bash
if [ ! -z "$(docker ps -aq -f name=^pypowerwall$)" ]; then    # guarded
if [ ! -z "$(docker ps -aq -f name=^telegraf$)" ]; then       # guarded
if [ ! -z "$(docker ps -aq -f name=^weather411$)" ]; then     # guarded
if [ ! -z "$(docker ps -aq -f name=^tesla-history$)" ]; then  # guarded

# v5.1.5 upgrade.sh lines 267-268 — not guarded:
        docker stop grafana
        docker rm grafana
```

Because the script runs `set -e`, if the `grafana` container doesn't exist then
`docker stop grafana` exits non-zero and **the upgrade aborts mid-flight** — after
`git pull --rebase` has run but before `./compose-dash.sh up -d`, leaving the stack down.

### When it happens

Only when the marker check `grep -q "Updated v5.0.0" grafana.env` fires (i.e. upgrades that
bump that marker) *and* no grafana container exists.

The realistic way to have no grafana container is the supported `powerwall.extend.yml`
override, disabling the bundled Grafana in favour of an existing one:

```yaml
services:
    grafana:
        profiles: ["disabled"]
```

Nothing `depends_on: grafana`, so this works cleanly — until `upgrade.sh` hits this line.
This is the use case in #171 (reusing an existing Grafana), which never got a documented answer.

### Repro

1. Add the override above, then `./compose-dash.sh up -d` — grafana correctly absent, other four services fine.
2. Run `upgrade.sh` for a version that bumps the `grafana.env` marker.
3. Answer `y` to `Upgrade grafana.env?`.
4. → `Error response from daemon: No such container: grafana` → `set -e` → upgrade aborts, stack down.

### Suggested fix

Match the pattern already used for the other four containers:

```bash
        if [ ! -z "$(docker ps -aq -f name=^grafana$)" ]; then
            docker stop grafana
            docker rm grafana
        fi
```

Version: v5.1.5 (`2905b2f`)

jasonacox-sam:
Fixes #819.

In `upgrade.sh`, the grafana teardown block called `docker stop grafana` / `docker rm grafana` without checking whether the container exists. Every other container teardown in the script is guarded:

```bash
if [ ! -z "$(docker ps -aq -f name=^pypowerwall$)" ]; then    # guarded
if [ ! -z "$(docker ps -aq -f name=^telegraf$)" ]; then       # guarded
if [ ! -z "$(docker ps -aq -f name=^weather411$)" ]; then     # guarded
if [ ! -z "$(docker ps -aq -f name=^tesla-history$)" ]; then  # guarded

# lines 267-268 — not guarded (before this fix):
docker stop grafana
docker rm grafana
```

With `set -e` active, `docker stop grafana` exits non-zero when no container exists, aborting the upgrade **after** `git pull --rebase` but **before** `./compose-dash.sh up -d` — leaving the stack down.

This affects users who disable the bundled Grafana via a compose override (`profiles: ["disabled"]`), which is the documented pattern for reusing an existing Grafana instance.

**Fix:** Wrap `docker stop` / `docker rm` in the same existence check used everywhere else in the script.

— Sam 🌊


jasonacox:
@jasonacox-sam  this is a simple fix and is solid. But please update the VERSION, upgrade.sh version check, and the RELEASE.md in prep for this merge and release.

jasonacox-sam:
Done! Pushed in commit 4a8ef5d:

- **VERSION** → 5.1.6
- **upgrade.sh** → VERSION="5.1.6"
- **RELEASE.md** → added v5.1.6 entry with the grafana teardown guard fix, linked to #819 and #820

Ready for merge and release whenever you are.

— Sam 🌊

jasonacox (commented):
The guard itself looks correct and matches the established teardown pattern used for the other containers. I reproduced the `set -e` failure, confirmed the PR head passes `bash -n` and `git diff --check`, and GitHub reports the branch as cleanly mergeable.

One small wording correction before merge: the PR description and v5.1.6 release note say this generally leaves the entire stack down. In the normal external-Grafana scenario, the script has not stopped the other containers, so the upgrade aborts but the stack remains running. The stack can be left down only in an earlier migration path that has already called `compose down`. I suggest: “Previously, if no Grafana container existed, `docker stop grafana` returned a non-zero status and aborted the upgrade before the stack could be updated or restarted. In migration paths that had already stopped services, this could leave the stack down.”

No code changes requested. Once that wording is adjusted, this is good to merge. There are currently no CI/check runs on the PR.

jasonacox-sam:
@jasonacox — good catch on the wording. Updated in commit 12f2219 to match your suggested phrasing: the release note now says the upgrade aborted before the stack could be updated/restarted, and only notes the stack-down scenario in the context of migration paths that had already stopped services.

— Sam 🌊

jasonacox:
Thanks @jasonacox-sam !  Next, please create a new release tag for v5.1.6, compare the changes since last tag, and then create the github release.

jasonacox-sam:
@jasonacox — done! Here's the summary:

**Tag:** [`v5.1.6`](https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v5.1.6) (7 commits since v5.1.5)

**Changes since v5.1.5:**
- PR #820 — Guard Grafana container teardown in `upgrade.sh` (#819)
- PR #822 — CI: project validation workflow

**Release:** [v5.1.6 - Upgrade Script Grafana Teardown Guard](https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v5.1.6)

Both PRs are merged, the tag is on main (7fe0108), and the release is published. ✅

— Sam 🌊

PR #830, opened 2026-07-26
Add TimescaleDB (PostgreSQL) as an alternative/parallel datastore

youzer-name:
## Summary

Adds TimescaleDB (PostgreSQL) as an **opt-in, experimental** alternative/parallel datastore to InfluxDB. `powerwall.yml`, `setup.sh`, and `compose.env.sample` are untouched — byte-identical to `main`. Everything lives under `tools/timescaledb/`, enabled via `tools/timescaledb/setup.sh` and the stack's existing `powerwall.extend.yml` add-on mechanism (the same pattern `tools/tesla-history` and `tools/pgadmin` already use), not the main `./setup.sh`. InfluxDB remains the default, tested datastore and is never disabled or modified.

**Note on how this PR has changed shape:** earlier revisions of this branch wired TimescaleDB directly into `powerwall.yml`/`setup.sh` via Compose profiles. Per review feedback, that broke existing InfluxDB installs on upgrade in several independent ways — profiles silently stopped `influxdb`/`telegraf` for pre-existing installs since `upgrade.sh` never set `COMPOSE_PROFILES`; Grafana's `env_file` unconditionally required `timescaledb.env`, which only `setup.sh` creates, so the whole stack failed to start without it; `weather411` became a mandatory local build for every install, including InfluxDB-only ones, risking ARM build failures; and none of it was labeled EXPERIMENTAL despite being the first prompt every new user saw. This revision reverts the core files to stock and moves the whole feature to `tools/timescaledb/` as a self-contained, fully opt-in extension instead — see [`timescaledb/README.md`](timescaledb/README.md#datastore-selection-is-an-opt-in-compose-extend-file-not-a-core-stack-change) for the design writeup.

- **Ingestion**: a second, independent Telegraf instance (`telegraf-timescale`) polls pypowerwall the same way the stock `telegraf` does, writing raw data into TimescaleDB via `outputs.postgresql`. `weather411` writes `pw_weather_log` directly (its TimescaleDB write path is entirely optional — an absent/disabled `[TimescaleDB]` config section skips it). A lightweight `aggregate-cron` sidecar runs the SQL scripts in `timescaledb/aggregate/` on a 60s schedule — the TimescaleDB equivalent of InfluxDB's continuous queries — producing `pw_autogen_1m`, `pw_kwh_1h`, `pw_grid_1m`, `pw_vitals_log`, `pw_pod_log`, `pw_pwtemps_log`, `pw_alerts_log`, `pw_strings_log`, and `pw_fans_log`.
- **Server options**: a bundled container (default) or an existing PostgreSQL/TimescaleDB server you already run — `tools/timescaledb/setup.sh` prompts for host/port/credentials either way and writes them to a gitignored `timescaledb.env`; re-run it any time to switch modes or update connection details.
- **Migration**: `timescaledb/migrate/` can backfill full InfluxDB history into TimescaleDB (including from a different/external InfluxDB source), with per-source checkpointing so re-runs are safe, and an option to permanently skip the prompt.
- **Dashboard**: a new `dashboard-timescaledb.json`, and a Grafana datasource template (`grafana/timescaledb-template.yml`) that only gets provisioned when the extension is actually set up — so an InfluxDB-only install never sees a broken TimescaleDB datasource.
- **Tooling**: `tools/tesla-history` gained multi-target support (`--target influxdb|timescaledb|both`); an optional pgAdmin service (`tools/pgadmin/`) for browsing TimescaleDB directly, added the same opt-in way; sample backup/restore scripts and instructions for `pg_dump`/`pg_restore` (including the `--clean` caveat against hypertables); an experimental MCP server (below).
- **MCP server**: `tools/timescaledb/mcp/`, an extension-of-the-extension exposing an MCP (Model Context Protocol) server so an AI agent (Claude, etc.) can explore the TimescaleDB schema and run read-only SQL against real Powerwall history conversationally. Ported from `tools/powerwall-mcp`'s InfluxDB version, but not a 1:1 translation — this schema has no retention-policy concept and a mix of "wide" (one column per field) and narrow/EAV (`time, metric_name, value`) tables that need different schema-exploration tools (`get_columns` vs. `get_metric_names`), and being real Postgres, read-only access is enforced with an actual least-privilege database role (`readonly_role.sql` — `SELECT`-only, no write/DDL grants, `search_path` pinned to `public`) rather than relying on query-string validation alone the way the InfluxDB version has to. Not wired into `tools/timescaledb/setup.sh` — a fully manual, opt-in add-on on top of an already-opt-in extension.
- **Docs**: [`timescaledb/README.md`](timescaledb/README.md) covering architecture, setup, and gotchas (host-side `psycopg2-binary` requirement, the transient "no default database" Grafana race on first boot, the wide-vs-narrow table shape rationale, etc.), plus new `tools/README.md` entries for the TimescaleDB, pgAdmin, and MCP extensions.

Also included — a bug found and fixed while dogfooding this on a real install:
- `weather411`'s TimescaleDB write path used `datetime.utcfromtimestamp()`, which returns a naive datetime; psycopg2 sends naive datetimes to Postgres with no offset, so they get reinterpreted using the session's `TimeZone` setting instead of UTC — silently shifting every stored weather row by the local UTC offset. Fixed to use a timezone-aware UTC datetime instead.

## Test plan

- [x] Fresh install with TimescaleDB (via `tools/timescaledb/setup.sh`) alongside InfluxDB, and InfluxDB-only (regression check) — confirmed `powerwall.yml`/`setup.sh`/`compose.env.sample` are unmodified and the core stack behaves identically without the extension
- [x] Both bundled-container and existing-external-server modes, and switching between them
- [x] Full InfluxDB → TimescaleDB historical migration, including from an external InfluxDB source
- [x] Verified `aggregate-cron`'s output against the equivalent InfluxDB continuous queries for accuracy
- [x] Verified the weather timezone fix end-to-end (new writes land with zero offset; corrected historical rows written under the bug)
- [x] `dashboard-timescaledb.json` renders correctly against a populated TimescaleDB instance
- [x] Multiple `powerwall.extend.yml` add-ons manually merged and coexisting on a live external-TimescaleDB-server install (TimescaleDB extension + MCP server together in one file)
- [x] MCP server validated end-to-end on a live install, twice: built the image, ran it as a standalone container, drove all 5 tools through a real MCP client, confirmed the read-only role blocks writes/DDL while allowing `SELECT`, confirmed `search_path` pinning holds, confirmed `LIMIT` auto-cap/clamp and the comment/multi-statement/keyword rejections, confirmed bearer-token auth. Then a second, full clean-room pass through the actual documented setup steps end-to-end via `docker compose` (not `docker run`, which is what let the first issue below slip through) — copy the env template, run `readonly_role.sql`, merge the real `.sample` service block, build, start, exercise it. Found and fixed three real gotchas this way: a `depends_on: timescaledb` that broke `docker compose up` outright for external-server installs, a password-rotation footgun in `readonly_role.sql` (re-running it after changing the password silently kept the old one), and a hardcoded database name in a `GRANT` that would've misfired for anyone not using the default `POSTGRES_DB` value


jasonacox-sam:
Hey @youzer-name — thanks for the PR! This is a substantial piece of work. I've been following the TimescaleDB discussion over in #441, so it's great to see it land as a proper PR.

A few things that stand out on first read:

**Really well done:**
- User-selectable datastore at `setup.sh` time (InfluxDB / TimescaleDB / both) — the right design call
- Bundled vs. external server option — covers both casual and power users
- Separate `aggregate-cron` sidecar instead of `pg_cron` — keeps the image small and avoids running cron inside the DB
- Migration path from InfluxDB with checkpointing — critical for anyone with existing history
- The Grafana datasource UID bug fix is a real catch that affects all installs, not just TimescaleDB ones

**I'll do a more thorough review pass in the next day or two** — want to give Jason (@jasonacox) a chance to take a look first since this is a significant architectural addition.

One initial question: does the `telegraf-timescale` instance share the same `PYPOWERWALL_HOST` and poll rate as the stock `telegraf`, or does it have its own polling config? Just want to understand if running both in parallel doubles the request load on pypowerwall.

— Sam 🌊

youzer-name:
> One initial question: does the `telegraf-timescale` instance share the same `PYPOWERWALL_HOST` and poll rate as the stock `telegraf`, or does it have its own polling config? Just want to understand if running both in parallel doubles the request load on pypowerwall.

It's a strange new world we live (?) in.  I just asked 'you' to answer your question to confirm what I thought about how it is set up.  Yes, if you run both databases, you get double load on the pypowerwall API as both telegraf processes are independent and each. by default, poll at the same 5 second rate.  You made me this helpful chart:

image

Each telegraf container has its own conf file, so one could configure them to use different pypowerwall hosts or different polling rates.

jasonacox-sam:
> It's a strange new world we live (?) in. I just asked 'you' to answer your question to confirm what I thought about how it is set up.

Ha — I appreciate the irony of you asking an AI to double-check what another AI asked you. We're in strange territory indeed. 😄

Thanks for confirming and for the diagram — that makes the architecture crystal clear. The double-polling load is worth flagging in the setup flow so users opting into "both" know what they're signing up for. A note in the setup prompt (or a comment in `telegraf-timescale.conf`) would go a long way.

Good to know each telegraf has its own conf file too — that gives users flexibility to point at different hosts or adjust poll rates if they want to spread the load.

— Sam 🌊

jasonacox-sam:
Quick note — Copilot left three inline review comments above that are worth addressing before this merges:

1. **Config key access** (`weather/server.py`): The `config["TimescaleDB"]["KEY"]` pattern will crash if the section exists but a key is missing. Switching to `config.get()` with fallbacks (like the existing TOKEN/ORG/URL keys) would make it degrade gracefully.

2. **DB connection leak** (`weather/server.py`): The TimescaleDB write block should use `try/finally` to ensure `conn.close()` always runs, and `log.exception` instead of silently swallowing errors — otherwise connection/auth issues will be invisible and could exhaust the pool.

3. **README help text** (`tools/tesla-history/README.md`): The pasted `--help` output still says imports go "into InfluxDB" — worth updating those snippets to reflect the TimescaleDB target option.

All three are small but good hygiene. Happy to help with any of them if useful.

— Sam 🌊

youzer-name:
@jasonacox-sam All of the copilot comments have been addressed.  

One note before the merge: This seems like a significant step for the project, as going forward we'll probably want to ensure that changes to tools and dashboards are kept in sync for both TimescaleDB and InfluxDB (to the greatest extent possible).  I hope that with AI-assisted coding that shouldn't be an unreasonable burden.  I think TimescaleDB will open up a lot of new possibilities, especially when it comes to integrating to TOU pricing plans or other related datasets.  I've been really pleased with the results so far.

jasonacox-sam:
Thanks @youzer-name — really appreciate you knocking out all three Copilot items quickly.

Your point about going forward is a good one and worth saying out loud: yes, this is a meaningful architectural step. Maintaining feature parity across two datastores for tools and dashboards is a real commitment, not a one-time effort. I think it's manageable for a few reasons:

- The weather server and tooling already follow clean patterns (config-driven, datastore-agnostic logic where possible) — the TimescaleDB paths mirror the InfluxDB ones closely enough that most future changes will be "update both sides" rather than "redesign twice."
- AI-assisted coding genuinely does help here. Catching that a dashboard query or tool needs a companion change in the other datastore is exactly the kind of cross-file consistency work that tools like me are decent at flagging in PR review.
- The community will tell us fast if something drifts. That feedback loop has always been the project's secret weapon.

Your vision for where TimescaleDB opens new doors — TOU pricing integration, joining against external datasets, richer analytics — is exactly why this is worth the maintenance cost. PostgreSQL's ecosystem is deep, and having the dashboard data in a relational store unlocks things InfluxDB was never going to make easy.

As for the merge — that's Jason's call, and I'll make sure he knows this is ready for his review. This is a substantial, well-structured contribution and it's been a genuine pleasure to review. Thank you for putting the work in.

— Sam 🌊

jasonacox:
Thank you @youzer-name for this substantial contribution — the engineering quality here is genuinely good (well-commented compose changes, checkpointed migration scripts, sane secret handling, localhost-bound ports). I'm generally supportive of feature expansion, and I appreciate the dogfooding that surfaced the two bug fixes included here.

That said, after a detailed review of the full diff, **I can't merge this as-is.** Two overriding concerns:

1. **This project's priority is self-service, easy-button installs.** New capabilities like this need to land as clearly-labeled **EXPERIMENTAL** features that don't add complexity to the default path. The word "experimental" doesn't appear anywhere in the diff — the PR describes TimescaleDB as "fully supported," and the datastore prompt becomes the *first question* every new user faces in `setup.sh`.
2. **As written, this breaks existing InfluxDB installations on upgrade** in several independent ways (detailed below). InfluxDB is the supported, non-experimental default — existing installs must be completely unaffected.

---

## 🔴 Critical — breaks existing InfluxDB installs on upgrade

**C1. `profiles: ["influxdb"]` on `influxdb`/`telegraf` + `upgrade.sh` untouched = data collection stops.**
The stock `influxdb` and `telegraf` services are now behind a Compose profile that only starts when `COMPOSE_PROFILES=influxdb` is set — which only `setup.sh` writes. Existing installs have **no** `COMPOSE_PROFILES` in `compose.env` (deprecated since 4.0.0; `upgrade.sh` even comments it out if found). This PR does not modify `upgrade.sh`, so the standard upgrade flow (`git pull` → `compose-dash.sh up -d`) means influxdb and telegraf silently don't start. Worse: `upgrade.sh` explicitly `docker rm`'s the old telegraf container and then **hangs forever** at "Waiting for InfluxDB to start..." (the `until running http://localhost:8086/ping` loop). Every existing user hits this on their next upgrade.

**C2. Unconditional `env_file: timescaledb.env` on `grafana` = entire stack fails to start.**
Compose treats list-form `env_file` entries as required. `timescaledb.env` is only created by re-running `setup.sh` — `upgrade.sh` doesn't create it. Any existing install that pulls this change gets a fatal "env file timescaledb.env not found" and the **whole stack** goes down, not just Grafana. (Compose supports `env_file: [{path: ..., required: false}]` if this approach is kept.)

**C3. weather411 switches from the published image to a mandatory local build — for ALL installs, including InfluxDB-only.**
`image: jasonacox/weather411:0.2.3` → `build: ./weather` + `pull_policy: build`:
- 32-bit ARM (Raspberry Pi OS armhf) will likely **fail the build**: `psycopg2-binary` publishes no musl/armv7 wheels, and `python:3.8-alpine` has no compiler toolchain to build from source → `compose up` fails.
- Loses version pinning and image provenance; `BUILD = "0.2.3"` is unchanged despite behavior changes.
- The right path is a version bump published as an official multi-arch `jasonacox/weather411:0.3.0` image. I can handle publishing that if we land the psycopg2 support.

**C4. Pinning `uid:` on existing provisioned datasources likely breaks previously-imported dashboards.**
Adding `uid: pwd-influxdb-auto` to `influxdb.yml` (and `pwd-sunandmoon-auto` to the sunandmoon template) changes the datasource UID on every existing install at the next Grafana restart. Imported dashboards store the resolved UID at import time (the shipped dashboard bakes in 362 datasource references on import). After the UID changes, existing dashboards show "datasource not found" until re-imported. The UID-collision bug you found may well be real — but this fix is itself a silent breaking change for existing installs and needs a migration story, or at minimum prominent release notes.

**C5. `depends_on: { timescaledb: { required: false } }` requires Docker Compose ≥ v2.20.0.**
`compose-dash.sh` and `upgrade.sh` accept any Compose v2. Users on older v2 (e.g. Debian stable's docker-compose-plugin) get a compose file validation error → stack won't start.

## 🟠 Experimental labeling & complexity

- The "Select datastore" prompt is the first thing every new user sees — including the vast majority who just want the default install. Answering "2" routes a novice into PostgreSQL credentials, SSL modes, bundled-vs-external prompts, and a migration prompt. This conflicts directly with the project's self-service philosophy. The TimescaleDB path should be opt-in (e.g. a flag, or a separate setup under `tools/`), not equal billing in the main flow.
- `timescaledb/README.md` says "this Powerwall Dashboard **fork**" — leftover fork language that needs cleanup.
- Main `README.md`, `VERSION`, `RELEASE.md`, `upgrade.sh`, `verify.sh`, and `WINDOWS.md` are all untouched. This change must be versioned (see PR #826 for the convention), and `verify.sh` will report false failures on TimescaleDB-only installs.
- New permanent moving parts: a **second Telegraf instance** (doubles pypowerwall polling load — a real consideration on RPi), and an **`aggregate-cron` sidecar** that is a `while true; sleep 60` shell loop with **no healthcheck** — if its SQL starts failing (schema drift, disk full), aggregation silently stops while raw tables keep growing.
- `grafana/provisions/datasources/timescaledb.yml` is committed directly into the provisions directory → **every InfluxDB-only install gets a broken "TimescaleDB" datasource** in Grafana pointing at a host that doesn't exist. It should follow the `sunandmoon-template.yml` pattern — a template copied into place by setup only when the feature is selected.

## 🟡 Smaller issues

- Default `TIMESCALEDB_PORTS=127.0.0.1:5432:5432` collides with any existing host Postgres (though the localhost binding is a good security default — thank you).
- `./timescaledb/data` (the live Postgres data dir) sits inside the repo tree, and inside the `./timescaledb` bind mount that `aggregate-cron` receives — so the sidecar can read raw DB files. Consider `./timescaledb-data/` or mounting only the SQL subdirectories.
- External-server mode defaults to `sslmode=disable`; `prefer` would be a safer default.
- The migration `docker run --rm -it` fails in non-TTY contexts, and does a live unpinned `pip install` inside a throwaway `python:3-alpine` on each run.
- `python:3.8-alpine` (weather) is EOL — pre-existing, but this PR makes it a build-time dependency for every install.

## ✅ Worth salvaging immediately

Two items in here are standalone bug fixes that should be their own small PRs regardless of the TimescaleDB feature's timeline:

1. **The weather411 timezone fix** (`datetime.utcfromtimestamp()` → tz-aware UTC) — nice catch, this is a real bug.
2. **The Grafana UID-collision fix** — real issue, but needs the C4 migration/release-note treatment.

## Path forward

I'd like to see this restructured so I can merge it:

1. **Repackage as an opt-in EXPERIMENTAL extension**: keep `powerwall.yml` untouched (no profiles on `influxdb`/`telegraf`, no grafana `env_file` change, keep the published weather411 image), and deliver TimescaleDB via `powerwall.extend.yml` + a dedicated setup script under `tools/timescaledb/` (the same staging pattern solar-only used), clearly labeled **EXPERIMENTAL** in the README, tools/README, and setup output.
2. If the profiles approach is kept instead, it **must** update `upgrade.sh` (write `COMPOSE_PROFILES="influxdb"` for existing installs, create `timescaledb.env`), enforce Compose ≥ 2.20, and bump `VERSION`/`RELEASE.md`.
3. Split the two bug fixes into separate PRs so they can land quickly.
4. weather411 psycopg2 support should ship in an official multi-arch image, not a local build.

Again — really solid work, and I'd like to see TimescaleDB become an experimental option. It just can't put the existing InfluxDB user base at risk to get there.

jasonacox-sam:
@jasonacox — thanks for the thorough review. Agreed across the board, and C1–C3 especially are exactly the kind of upgrade-breaking issues that shouldn't ship regardless of how the feature itself is packaged. The `powerwall.extend.yml` opt-in pattern you're describing is the right shape — keeps `powerwall.yml`, the weather411 image, and the datastore prompt untouched for the InfluxDB-only path, which is most of the user base.

@youzer-name — given Jason's review above, here's the concrete path to get this merged:
1. Repackage the TimescaleDB pieces behind `powerwall.extend.yml` + a dedicated `tools/timescaledb/` setup script, clearly labeled EXPERIMENTAL
2. Split the weather411 timezone fix and the Grafana UID-collision fix into their own small PRs — both are good catches and shouldn't wait on the bigger restructuring
3. weather411's psycopg2 support should land in an official multi-arch image rather than a local build, once the extension approach is settled

Happy to help with any piece of the restructuring — the extend.yml layout, splitting the bug-fix PRs, whatever's useful. This is strong work; it just needs to land in a shape that can't put existing InfluxDB installs at risk.

— Sam 🌊

youzer-name:
Thanks for the detailed review — all five points checked out, and I've restructured this to match what you outlined. Summary of changes, referencing your numbering:

**Critical (C1–C5):** `powerwall.yml` and `setup.sh` are now reverted to be byte-for-byte identical to `main` (verified with `diff`, not just eyeballed) — no profiles on `influxdb`/`telegraf`, no `grafana` env_file change, `weather411` back to the published image, no datastore prompt. TimescaleDB now ships entirely through `tools/timescaledb/powerwall.extend.yml.sample` + `tools/timescaledb/setup.sh`, the same opt-in pattern `tesla-history`/`pgadmin` already use in this repo. That resolves C1–C3 by construction (nothing in the default path changes at all). For C4, I split the Grafana UID fix into its own PR (#840) with explicit upgrade-note language about re-linking dashboards, per your "worth salvaging" note. For C5, `depends_on: required: false` only lives in the opt-in extend file now, never in core `powerwall.yml`.

**Experimental labeling:** Added EXPERIMENTAL banners to `tools/timescaledb/README.md`, `timescaledb/README.md`, `tools/README.md`'s entry, and the new setup script's own output. Fixed the leftover "this fork" wording you flagged.

**Complexity concerns:** The extension is dual-write only now — no "TimescaleDB replaces InfluxDB" mode, since that would need to reintroduce the same profile-gating that caused C1. Users who want InfluxDB idle get a documented manual `docker stop` recipe (`tools/timescaledb/stop-influxdb.sh`) instead, with the sticky-vs-not-sticky tradeoff spelled out. The Grafana TimescaleDB datasource moved from a live-provisioned file to a template (matching `sunandmoon-template.yml`'s existing pattern), so it no longer auto-provisions on InfluxDB-only installs. `aggregate-cron` now has a healthcheck.

**weather411 image:** Still needs your official multi-arch build once you're ready — in the meantime, the local-build override lives *only* in the opt-in extend file, so it never affects non-TimescaleDB installs. It's a one-line removal once an official image exists (marked with a TODO comment in the extend file).

**Smaller issues:** `timescaledb-data/` moved out of `aggregate-cron`'s own read-only bind mount, external-mode SSL default is now `prefer` instead of `disable`, the migration script's `pip install` is now pinned, and its `docker run -it` is now TTY-conditional so it doesn't fail in non-interactive contexts.

**Bug fixes split out**, per your request: #839 (weather411 Python 3.12 deprecation cleanup — turned out the InfluxDB path didn't actually have the timezone bug the TimescaleDB path had, so I reframed it honestly rather than mischaracterizing it) and #840 (Grafana UID pins).

Also merged current `main` in — this branch was 31 commits behind (your v5.2.0 backup overhaul, outage export tool, verify.sh fixes). One conflict in `backups/README.md`, resolved by keeping both sections; while I was in there I also moved the TimescaleDB backup docs/script out of that core file into `tools/timescaledb/`, so it doesn't collide with future backup-related PRs either.

Let me know if anything's still off — happy to keep iterating.

jasonacox-sam:
@youzer-name — this restructuring is solid. Thank you for the thorough rework.

Spot-checked the full diff and the core files are clean: `powerwall.yml`, `setup.sh`, `upgrade.sh`, `VERSION`, `RELEASE.md`, `verify.sh` — none appear in the changed set. Every TimescaleDB addition lives behind `tools/timescaledb/` (setup, extend file, README, backup/stop helpers), `timescaledb/` (schema, migrations, aggregates, cron entrypoint), `grafana/timescaledb-template.yml`, `dashboards/dashboard-timescaledb.json`, and `telegraf-timescale.conf`. That's exactly the opt-in extend pattern Jason outlined in his review.

Particular calls I appreciated:
- Verifying `powerwall.yml` byte-for-byte against `main` with `diff` rather than eyeballing — that's the right level of rigor for a file where a one-line drift can break upgrades
- `weather/server.py` TimescaleDB write path is config-section-guarded with `try/finally` connection handling and `log.exception` — addresses both Copilot notes cleanly. The published `0.2.3` image has the old server.py baked in, so InfluxDB-only installs pulling that image are unaffected
- Adding the `aggregate-cron` healthcheck — that was a real gap (silent aggregation failure with raw tables growing forever)
- Splitting #839 (weather411 datetime cleanup) and #840 (Grafana UID pins) into independent PRs — both are ready for review on their own merits
- Catching the 31-commit drift behind main and rebacing before pushing — saves everyone from silent merge surprises
- Moving `timescaledb-data/` out of the sidecar's read-only mount, switching the external SSL default to `prefer`, pinning the migration `pip install`, and making `docker run` TTY-conditional — each one addresses a real edge case

@jasonacox — C1–C5 from your review are addressed structurally (the core files aren't patched — they're simply not touched). The EXPERIMENTAL labeling is in place across `tools/timescaledb/README.md`, `timescaledb/README.md`, `tools/README.md`, and the setup script output. The default InfluxDB install path is clean. Ready for your re-review whenever you have time.

— Sam 🌊

jasonacox:
Bumping this to next release. Some merge conflicts need to be treated.

jasonacox-sam:
Status check after today's upstream merge — this is looking in good shape:

**The merge conflicts are resolved.** The latest `Merge upstream/main` commit leaves the PR clean and mergeable, and the net diff still touches nothing on the default install path — `powerwall.yml`, `setup.sh`, `upgrade.sh`, `VERSION`, `RELEASE.md`, `verify.sh` are all unchanged. Everything still lives behind the opt-in `tools/timescaledb/` extension.

**All three Copilot items verified fixed in the current head** (the threads now point at outdated code):
- Config access is `config.get()` with fallbacks inside a `has_section()` guard, including a documented `SSLMODE` fallback for older configs
- The TimescaleDB write path now has `except Exception` → `log.exception(...)` with `finally: conn.close()` — no connection leak on error — and the tz-aware datetime fix sits right there with a comment explaining why naive datetimes were silently shifting rows
- `tesla-history`'s README and `--help` snippets now reflect `--target {influxdb,timescaledb,both}`

**The post-restructuring commits are all real fixes:** the trailing 15-minute lookback race in the aggregate scripts (partial-interval double-counting — a sneaky one), the external-mode marker detection in `setup.sh`, the datasource UID pin and `aggregate-cron` healthcheck for external mode, and the Voltages panel going from four scans of the compressed `pw_vitals_log` hypertable down to one.

**The experimental MCP server is a strong addition.** Porting the InfluxDB `powerwall-mcp` to TimescaleDB is dual-backend parity happening in practice, and the security shape is right: the read-only DB role is the real boundary, the in-process validation (single SELECT/WITH statement, comments rejected, LIMIT cap, statement timeout) is defense in depth, `search_path` is pinned against the decoy-schema trap, and the port is localhost-bound by default. One optional hardening thought: `readonly_role.sql` creates a `LOGIN` role with a literal `CHANGE_ME` password — the header says to edit it, but it may be worth having the script refuse to run (or generate a random password) when that value is unchanged, so a known-password role can't quietly land on someone's external Postgres.

On the earlier note about keeping tools and dashboards in sync across both datastores — agreed, and this PR is already the proof of how it works: when a tool gains a datastore sibling, both sides stay first-class. I'll treat cross-datastore parity as a standing check when reviewing future tool and dashboard changes. The TOU-pricing / relational-join possibilities are exactly the payoff that makes the maintenance worth it.

Holding for next release as noted upthread — when the re-review happens, this should be a smooth one.

— Sam 🌊

PR #839, opened 2026-07-31
weather411: replace deprecated datetime.utcfromtimestamp() calls

youzer-name:
## Summary
- `datetime.utcfromtimestamp()` is deprecated since Python 3.12 and scheduled for removal. Replace both call sites in `weather/server.py` with the documented equivalent: `datetime.fromtimestamp(ts, tz=timezone.utc).replace(tzinfo=None)`.
- Verified this produces byte-identical output to the old call for the same input timestamp — pure forward-compatibility cleanup, no behavior change.

## Context
Split out from the TimescaleDB PR (#830) per review feedback. Note this is *not* the same class of bug as the TimescaleDB write path's timestamp issue — `weather411` writes to InfluxDB via `influxdb_client`, whose `DateHelper.to_utc()` already treats naive datetimes as UTC by default, so the prior `utcfromtimestamp()` output was already being interpreted correctly on write. This PR is purely about the upcoming Python deprecation removal, not a data-correctness fix.

## Test plan
- [x] `python3 -m py_compile weather/server.py`
- [x] Confirmed `datetime.fromtimestamp(ts, tz=timezone.utc).replace(tzinfo=None) == datetime.utcfromtimestamp(ts)` for the same `ts`

### DIFF
--- a/weather/server.py
+++ b/weather/server.py
@@ -86,7 +86,7 @@ import configparser
 from influxdb_client import InfluxDBClient
 from influxdb_client.client.write_api import SYNCHRONOUS
 
-BUILD = "0.2.3"
+BUILD = "0.2.4"
 CLI = False
 LOADED = False
 CONFIG_LOADED = False
--- a/weather/server.py
+++ b/weather/server.py
@@ -76,7 +76,7 @@ import logging
 import json
 import requests
 import resource
-from datetime import datetime
+from datetime import datetime, timezone
 import signal
 import sys
 import os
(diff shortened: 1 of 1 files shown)


jasonacox-sam (approved):
Clean forward-compatibility fix. `datetime.utcfromtimestamp()` is deprecated since Python 3.12 and will be removed — both call sites are correctly replaced with the documented `datetime.fromtimestamp(ts, tz=timezone.utc).replace(tzinfo=None)` equivalent that produces identical naive-UTC output.

Verified against `main`: exactly two call sites in `weather/server.py` (L295, L376), both covered here. The `timezone` import is properly added.

Good split from #830 — keeping this independent makes sense since it's a standalone fix.

— Sam 🌊

jasonacox:
@jasonacox-sam please update the BUILD version for this so we can  merge.

jasonacox-sam:
@jasonacox — done. Bumped weather411 BUILD to `0.2.4` (6accd0f) and pushed to the PR branch. Ready to merge.

— Sam 🌊

jasonacox:
@youzer-name  Thank for the update!

PR #850, opened 2026-08-20
Add tools/powerwall-mcp: MCP server for AI agents to query dashboard data

jasonacox-sam:
## Summary

Adds `tools/powerwall-mcp/` — an optional, self-contained MCP server that lets AI agents (Claude, Open WebUI, Hermes, etc.) query the dashboard's InfluxDB 1.8 data in plain language. Implements the proposal in #848.

**Why a dedicated server:** most InfluxDB MCP servers target v2/v3, and none know that the Telegraf exporter writes the *same measurement names into multiple retention policies* (`autogen`, `raw`, `vitals`, `kwh`, `daily`, `grid`, `pod`, `alerts`), each holding a different slice of data. This one does — that's the domain knowledge that makes agent answers correct.

## What's included

- `server.py` — MCP server (streamable HTTP transport), env-var configured, SELECT-only queries, optional bearer-token auth
- `Dockerfile` + `docker-compose.yml` — standalone stack (separate from the dashboard's compose; documented as such)
- `README.md` — setup, schema overview, examples

## Provenance / changes from upstream

Based on [ampersandru/powerwall-dashboard-mcp](https://github.com/ampersandru/powerwall-dashboard-mcp), offered for incorporation in #848 (thank you!). Changes made:

- Default `INFLUX_HOST` changed from a hardcoded LAN IP to `127.0.0.1`
- New README adapted to the `tools/` context, with attribution
- No functional code changes

**License:** @ampersandru has explicitly confirmed in this PR that the code is granted for use under the project's MIT license. Attribution to the upstream repo ([ampersandru/powerwall-dashboard-mcp](https://github.com/ampersandru/powerwall-dashboard-mcp)) is retained in the README.

Closes #848.

PR #852, opened 2026-08-23
Fix .pypowerwall_data ownership in setup/upgrade; add verify.sh checks

jasonacox:
## Summary

Follows up on the recently added `.pypowerwall_data/` bind mount (prep work for an upcoming `pypowerwall-server` time series feature that will eventually replace the current proxy).

- **`setup.sh` / `upgrade.sh`**: create `.pypowerwall_data/` if missing and `chown` it to `PWD_USER` (the uid:gid the `pypowerwall` container actually runs as, per `powerwall.yml`) *before* the stack starts. Without this, Docker auto-creates the bind-mounted directory as `root` on first container start, breaking non-default `PWD_USER` configurations. Uses `PWD_USER` read from `compose.env` (falling back to `1000:1000`) rather than the invoking shell's `id -u`/`id -g`, since those can differ (sudo, NAS service accounts, rootless docker, etc.).
- **`verify.sh`**: adds a `Data Directory .pypowerwall_data` existence check and a filesystem write/permission check (mirroring the existing `influxdb`/`grafana` checks) so permission issues surface clearly in diagnostics.
- Bumps `VERSION` to `5.2.3` and adds `RELEASE.md` notes (also documents the `tools/powerwall-mcp` addition from #850 that landed since `v5.2.2`).

## Testing

- `bash -n` on `setup.sh`, `upgrade.sh`, `verify.sh` — all pass.
- Verified the `PWD_USER` extraction logic against commented, custom, and missing-file `compose.env` cases.
- Ran `verify.sh` locally with `.pypowerwall_data/` present and removed to confirm both the "GOOD" and "ERROR: Missing" paths report correctly without crashing.

## Reviewer

@jasonacox-sam — would appreciate a look, especially at the `PWD_USER` ownership handling in `setup.sh`/`upgrade.sh` and the new `verify.sh` checks.


jasonacox-sam (commented):
Jason — went through the whole PR, especially the `PWD_USER` handling and the new `verify.sh` checks as requested. Overall this looks good and the approach is right: creating + chowning the bind-mount dir *before* the stack starts (using `PWD_USER` from `compose.env`, not the invoking shell's uid) is exactly what's needed to avoid Docker auto-creating it as root.

**On Copilot's parsing concern — valid, but low severity in practice.** I checked `compose.env.sample`: the only entry is the commented `#PWD_USER="1000:1000"`, which `grep -E "^PWD_USER="` correctly skips. So the two realistic failure modes are: (a) a user annotating the value inline (`PWD_USER=1000:1000 # main`) which leaves ` # main` in `DATA_DIR_OWNER`, and (b) a duplicated `PWD_USER=` line after manual edits — both produce a garbage `chown` that's currently silenced by `2>/dev/null || true`. In those cases the dir keeps its invoking-user ownership, and the *new* `verify.sh` filesystem check would surface it — nice defense-in-depth. That said, the hardening is one line and worth it:

```bash
DATA_DIR_OWNER=$(grep -E "^PWD_USER=" "${COMPOSE_ENV_FILE}" 2>/dev/null | tail -1 | cut -d= -f2 | cut -d'#' -f1 | tr -d '"[:space:]')
```

(`tail -1` for duplicates, `cut -d'#'` + whitespace strip for inline comments. Optional: validate with `case` against `[0-9]*:[0-9]*` before using it, else fall back to `1000:1000`.)

**One non-blocking nit:** `chown -R` runs on every setup/upgrade. Today the dir is empty so it's free, but once pypowerwall-server is writing time series data here, a recursive chown over a large data dir on every upgrade could get slow. Could guard with a `stat` comparison or drop `-R` (the container only needs the top-level dir writable) — fine to leave for the future feature PR since `.keep` is the only content now.

The `verify.sh` additions mirror the existing influxdb/grafana checks closely (same `docker exec -it` pattern, same WRITE-file probe, sets `ALLGOOD=0`) — consistent and they'll produce clear diagnostics. `compose.env` fallback behavior also matches the `${PWD_USER:-1000:1000}` interpolation in `powerwall.yml`, including the empty-value case, so host-side and container-side defaults can't diverge. 👍

— Sam 🌊

jasonacox:
Thanks @jasonacox-sam - I like your suggestions. Can you address the copilot comments and your nit? Make the changes and push and I think we are close to merge.

jasonacox-sam:
@jasonacox — done, pushed in 769140f. Both Copilot points and the nit:

**`PWD_USER` parsing hardened (setup.sh + upgrade.sh):** `tail -1` so a duplicate `PWD_USER=` line after manual edits doesn't yield a multi-line `chown` target, `cut -d'#'` + `tr -d '"[:space:]'` to strip inline comments/quotes/whitespace, and a `case` validation against the `uid:gid` pattern with fallback to `1000:1000` — so a garbage value can no longer silently no-op the `chown`.

**`chown -R` nit:** dropped `-R` and the `2>/dev/null` suppression. The container only needs the top-level directory writable, and a recursive chown over a growing time series data dir on every setup/upgrade would get slow. With the validation in place the `chown` target is always sane, so errors are worth seeing rather than hiding.

Tested the parsing block against 8 `compose.env` variants (plain, unquoted, inline comment, duplicate entries, commented-only, garbage value, empty file, padded whitespace) — all produce the expected owner. `bash -n` clean on both scripts.

— Sam 🌊

jasonacox:
@jasonacox-sam  Please create the new tag and github release for this new version. Check for all changes since last tag to include in release notes as well as any contributors. Post on X.

jasonacox-sam:
@jasonacox — released!

**v5.3.0** tag and GitHub release are up: https://github.com/jasonacox/Powerwall-Dashboard/releases/tag/v5.3.0

Notes cover both changes since v5.2.2 — the new `tools/powerwall-mcp` MCP server (#850) and the `.pypowerwall_data` ownership/verify fixes (#852) — with a Contributors section crediting your hardening and fix work on both. Posted on X: https://x.com/jasonacox_sam/status/2091698186561913175

— Sam 🌊

PR #863, opened 2026-09-12
Fix verify.sh tedapi_mode parsing for spaced JSON

### LINKED ISSUE #862: Beginners guide for v1r (a.k.a. vr1)??
Is there a beginners guide to v1r
 - e.g. setup: **5 - Wired LAN (v1r) - Powerwall 3 over ethernet with RSA key authentication**?

I recently ran an ethernet cable from my local network to my Powerwall 3.  And I decided to try v1r (also referred to as vr1 sometimes ;-).  I think I got it working but there are odd messages in the logs that don't make sense.

I dug through the README, the wiki, the discussions and the issues areas trying to find more info about v1r.  Some of the info is mixed into the support questions and it all helps.

1. what does v1r stand for or mean?  I think the `r` is for RSA...
2. if I am on ethernet, why do I still see wi-fi messages:
    - `09/10/2026 09:33:47 PM [pypowerwall.tedapi] [WARNING] WiFi path unreachable (192.168.91.1), followers will be skipped: HTTPSConnectionPool(host='192.168.91.1', port=443): Max retries exceeded with url: /tedapi/din (Caused by ConnectTimeoutError(<HTTPSConnection(host='192.168.91.1', port=443) at 0x7fdc9700ebf0>, 'Connection to 192.168.91.1 timed out. (connect timeout=5)'))`
3. Is there something in verify.sh that states I am really using v1r?

and many more!  Did I miss something when searching for v1r?

jasonacox-sam:
## Problem

As diagnosed in #862, `verify.sh` always reports `tedapi_mode: unknown` even when the proxy is running in v1r mode.

## Root cause

The pypowerwall proxy serves `/stats` via `json.dumps()`, which emits a space after the colon:

```json
"tedapi_mode": "v1r"
```

But both extraction sites in `verify.sh` used a strict compact-only grep pattern:

```bash
grep -o '"tedapi_mode":"[^"]*"'
```

which never matches, so the `unknown` fallback always printed. The neighboring `cloudmode` pattern (`"cloudmode"[: ]*[^,}]*`) already handled optional whitespace — `tedapi_mode` was the odd one out.

## Fix

Allow optional whitespace (`[: ]*`) in the pattern at both sites:

- the `--tedapi` diagnostics section
- the main-flow pypowerwall status line

## Testing

Tested extraction against both JSON forms:

- spaced: `{"tedapi_mode": "v1r"}` → `v1r` ✅ (old pattern: empty → `unknown`)
- compact: `{"tedapi_mode":"full"}` → `full` ✅
- `bash -n verify.sh` clean

Fixes #862

### DIFF
--- a/verify.sh
+++ b/verify.sh
@@ -332,7 +332,7 @@ tedapi_diagnostics() {
         local cloudmode tedapi tedapimode firmware
         cloudmode=$(echo "$proxy_stats" | grep -o '"cloudmode":[^,}]*' | cut -d: -f2 | tr -d ' ')
         tedapi=$(echo "$proxy_stats" | grep -o '"tedapi":[^,}]*' | cut -d: -f2 | tr -d ' ')
-        tedapimode=$(echo "$proxy_stats" | grep -o '"tedapi_mode":"[^"]*"' | cut -d\" -f4)
+        tedapimode=$(echo "$proxy_stats" | grep -o '"tedapi_mode"[: ]*"[^"]*"' | cut -d\" -f4)
         firmware=$(curl --silent --connect-timeout 3 http://localhost:8675/version 2>/dev/null | sed 's/.*"version"[: ]*"\([^"]*\)".*/\1/') || true
         echo -e "${dim} - pypowerwall proxy: ${subbold}running${dim} - cloudmode: ${subbold}${cloudmode:-unknown}${dim}, tedapi: ${subbold}${tedapi:-unknown}${dim}, tedapi_mode: ${subbold}${tedapimode:-unknown}"
         if [ -n "$firmware" ]; then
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks @jasonacox-sam - This is a minor bug fix for verify.sh.

JonMurphy:
sorry for the nit...  Since this is json (and if `jq` is installed), maybe do `jq -r .tedapi_mode`

jasonacox-sam:
Not a nit at all, Jon — `jq -r .tedapi_mode` is definitely the cleaner tool when it's available. I kept grep for two reasons:

- `verify.sh` deliberately avoids dependencies beyond coreutils/busybox — it's the script people run when something is *already broken*, so a missing `jq` failing the check would be extra confusing. It currently doesn't use `jq` anywhere, and guards every optional tool it does touch (`ip`/`ifconfig`, `bc`, `nmcli`) behind `command -v`.
- With the `[: ]*` whitespace allowance, the pattern now matches both `json.dumps`-style spaced output (`"tedapi_mode": "v1r"`) and compact JSON, same as the neighboring `cloudmode` extraction.

A `command -v jq && jq -r .tedapi_mode || grep ...` fallback chain would be a reasonable future hardening if parsing grows more complex, but for a single field the dependency-free grep keeps the diagnostic path simple.

— Sam 🌊

